PR #41, opened 2015-01-19
CONTRIB-4696 Make CI exclusions aware of distributed thirdpartylibs.xml 

stronk7:
While the local_codechecker already had support for the exclusion of thirdpartylibs... CI jobs (define_excluded) was missing it and instead we were using a fixed list.

With these changes (already tested by David @ laptop), all CI jobs will be able to get a better list of excluded stuff, automatically, not requiring changes to the fixed list but getting paths from thirdpartylibs information.

The laptop has been also running the usual CI jobs for the last 5-6 hours with this patch applied, apparently without a problem.

https://tracker.moodle.org/browse/CONTRIB-4696

Ciao :-)

### DIFF
--- a/define_excluded/define_excluded.sh
+++ b/define_excluded/define_excluded.sh
@@ -79,6 +79,26 @@ yui/build/
 *.png
 *.svg"
 
+# Now, look for all the thirdpartylibs.xml in codebase, adding
+# all the found locations to the list of excluded.
+if [[ -n ${gitdir} ]]; then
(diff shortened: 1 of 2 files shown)


stronk7:
Thanks for your awesome work, Eloy!

stronk7:
Thanks Eloy, much appreciated!

danpoltawski:
:gun:

stronk7:
ROFL!

PR #42, opened 2015-02-02
MDLSITE-3799 ci: reset fixfor versions when moving out from integration.

stronk7:
Reopened issues that have failed in the testing phase, already have got
their fixfor versions set. Once they are reopened and moved out from
current integration cycle those fixfor versions, now incorrect, remain
set.

This change just ensures that any issue moved out from integration
(reopened) gets the versions cleaned.


danpoltawski:
Nice - merged - thanks

PR #43, opened 2015-03-30
MDLSITE-2892: Guarantee no null priority lands to integration

stronk7:
This implements #4 of https://tracker.moodle.org/browse/MDLSITE-2892
because it's leading to wrong orders in the integration process.

The rest can wait for some time, but this one is hiting us with
old issues having null altering the expected (0..1) order.


danpoltawski:
Seemed to work perfectly and was straight forward enough

stronk7:
thanks!

PR #44, opened 2015-04-02
MDLSITE-2317 first cut of autorebaser

danpoltawski:
I considered naming this more generically than 'security rebaser' - but in the end I decided to keep the terminology focused on security so its easier to understand.

The script does not attempt to auto-setup the security repository, so the first step to use this is to setup the required branches in the repo to match integration.git (or moodle.git, doesn't really matter):

i.e. for master:
integration/master == security/security-master == security/upstream-ref-maser

From that point on, integrators can start pulling to security-master and it'll start to autorebase.


danpoltawski:
I have merged this into the laptop branch for testing.

stronk7:
Sweet, and seems to be working... let's see how it proceeds when we start to send stuff to integration.git

danpoltawski:
Am self merging as we kinda agreed in integration chat.

PR #45, opened 2015-04-28
MDLSITE-4018 bulk-checker: process all awaiting integration issues

stronk7:
Before this, only the oned under current integration were
being checked and that was leading, on continuous periods,
where everything happens quickly to some issues never checked.

With this change we'll start to process any issue awaiting integration
that never has been checked before (missing ci label).


stronk7:
closed to make things easier to zulus

PR #47, opened 2015-05-03
MDLSITE-4023 - Various fixes

stronk7:
This includes 3 commits, all them about https://tracker.moodle.org/browse/MDLSITE-4023:

1) Cleanup of errors.txt output. So all messages will include a level (error, warn, info).
2) Implement gc (normal and aggressive) after a given number of executions (25 and 900 by default). Also, amend the "find" commands so they always, explicitly, exclude .git (and work).
3) When the pre-checker ends with error, look explicitly for controlled "error level" messages. If present, show them. If missing, it's an uncontrolled message and special output is generated, suggesting to the dev to create an issue/contact with integrator.

Ciao :-)


stronk7:
Done, added new commit for handling csslint problems as official errors.

PR #48, opened 2015-05-07
MDLSITE-3899 Stop running jshint on built folders

danpoltawski:
The problem was the existing rule relied on having subdirectories (i.e.
lib/yui/build/some-random-module/built.js for yui). Where as amd doesn't
have this strcuture so its just lib/amd/build/some-module-min.js). Doing
it this way means a more blanket ignore for yui, but think thats
desierable anyway.

Branch to test this with:
git://github.com/danpoltawski/moodle.git jshinttest

(Test branch Includes 3 commits off master which change yui/amd and normal srv/build direcories)


danpoltawski:
(Doh, there is unrelated commit from Eloy there)

stronk7:
My commit deleted. had it there to verify phpunit continued working ok with it.

stronk7:
thanks! I did look there days ago and was unable to understand why it was not working, blind me!

PR #49, opened 2015-05-11
MDLSITE-4044 prechecker: create local branches when valid and needed

stronk7:
This usually happens only on first run after a major release, where
new branches for patches are available.

Note that for public server I already created the branch manually, to get CiBoT working ASAP. So it won't create anything there.

So I've used the laptop to implement and test this, and seems to be working ok (see the new infos about branch being created):

http://ci.stronk7.com/view/prehecker/job/Precheck%20remote%20branch/1566/console

With this accepted, new future branches MOODLE_30_STABLE.... will be created automatically in the CIs clones.

Ciao :-)


danpoltawski:
Merged, thanks!

PR #50, opened 2015-06-30
NOBUG: Avoid weird dereferenced links std behavior.

stronk7:
By default, ln itself follows links so, at the end
it's highly possible to get them recursively created
no matter they already exists. To solve this the -n
flag can be used, causing ln not to follow links.

Or, something like that, really I've not thought too
much about it. I just tried the flag and verified it
solved the problem, so applying it everywhere.

Ref: http://www.gossamer-threads.com/lists/linux/kernel/1966

### DIFF
--- a/grunt_process/grunt_process.sh
+++ b/grunt_process/grunt_process.sh
@@ -56,7 +56,7 @@ else
 fi
 
 # Linking it.
-ln -fs ${npmbase}/${gitbranch}/node_modules ${gitdir}/node_modules
+ln -nfs ${npmbase}/${gitbranch}/node_modules ${gitdir}/node_modules
 
 # Verify there is a grunt executable available, installing everrything if missing
 if [[ ! -f ${gitdir}/node_modules/grunt-cli/bin/grunt ]]; then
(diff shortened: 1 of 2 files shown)


danpoltawski:
Shouldn't we check if it already exists?

stronk7:
think composer one, always deletes and recreates it, that's the cause there weren't problems there. In the other side, grunt was not, because I was using the -f (sort of force) flag that I discovered recently.

Problem is that dereferencing takes sort of precedence over -f, so it was causing the 2nd link to be created within the directory. That's the reason -n comes to the rescue, preventing the deferencing to happen, so -f is able to "recreate" (delete and create) the link.

Basically "ln -snf" === "rm & ln -s". While looking for an explanation, I found lots of pages talking about "ln" dereferencing and in many of them the solution was, simply, to always use "ln -snf" and forget.

Ciao :-)

stronk7:
Another ref: http://superuser.com/questions/645842/how-to-overwrite-symblic-link-of-directory

danpoltawski:
Thanks for explaining - merging..

PR #53, opened 2015-10-10
MDLSITE-4198 ci: split maxcommits into warning(10) and error(100)

stronk7:
To better support branches with a number of commits, the old, unique,
maxcommits setting is changed plugin-wide to use a couple of settings
instead.

Also, we are moving from an eroor&stop behavior (where the rest of the
checks were not applied) to a error/warn and continue behavior, by
showing the maxcommits warnings and errors within the commits checker.

This may lead to problems is somebody does submit a huge/wrong branch,
but that also was happening right now irrespectively of the # of
commits. If we need countermeasures for that situation, time will tell.

### DIFF
--- a/tracker_automations/set_integration_priority_to_one/set_integration_priority_to_one.sh
+++ b/tracker_automations/set_integration_priority_to_one/set_integration_priority_to_one.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+# Look for issues known to need their integration priority raised.
+#jiraclicmd: fill execution path of the jira cli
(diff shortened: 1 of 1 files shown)


danpoltawski:
the pull request has two commits now, but +1 to merge the maxcommtis one.

stronk7:
I'm closing this pull request and creating new one pointing to own branch instead of "laptop" one. Sorry for the noise.

stronk7:
#54 was created to finally pull this stuff, FYI.

PR #55, opened 2015-10-12
MDLSITE-4200 ci: raise integration priority when needed

stronk7:
There are some well known situations:
  - issue has "mdlqa" label.
  - issue has certain "Must fix for X.Y" version.

Where we must give those issues higher prioring under
continuous integration periods. Till now we do that using
separate filters or custom searches.

Instead, with this job, those issues will be searched periodically
and their "integration priority" will be raised to 1, so they will
appear the first under unmodified pickers.


danpoltawski:
Merged thanks.

PR #56, opened 2015-10-21
MDLSITE-1973 versions: Verify wrong and future dates are not allowed

stronk7:
While this does not complete all the checks to be implemented @
MDLSITE-1973, the 2 new tests added:

- Verify that the YYYYMMDD digits are correct, existing date.
- Verify that the YYYYMMDD digits are actual (not future >7d) date.

Do exactly match the 2-3 latest problems we have had lately. So
the earlier we are checking for them, the better.

### DIFF
--- a/versions_check_set/versions_check_set.sh
+++ b/versions_check_set/versions_check_set.sh
@@ -102,6 +102,31 @@ for i in ${allfiles}; do
         fi
     fi
 
+    # Verify the version 8 first digits do lead to a valid YYYYMMDD (only if we have version)
+    if [ ! -z "${version}" ]; then
+        # Extract 8 first digits (we already know they are there)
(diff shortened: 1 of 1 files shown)


stronk7:
I've been running it here leading to these new lines when the tests are ok:

```
- mod/glossary/version.php:
  ...
  + INFO: Correct version first 8 digits date (2015-06-02)
  + INFO: Correct actual (<+7d) date found (2015-06-02)
```

And with error:
```
- mod/glossary/version.php:
  ...
  + ERROR: No correct version first 8 digits date (date: invalid date «20150533»)
  + ERROR: No correct actual (<+7d) date found (2016-06-02)
```

danpoltawski:
Nice - thanks Eloy - merged

PR #59, opened 2015-11-06
MDLSITE-3701 tracker: query issues 'peer review in progress' for checks

danpoltawski:
If issues are peer reviewed too quickly, they do not get the ci label. 

*note* I haven't tested this yet ;-)


stronk7:
np here. Cannot imagine it causing much trouble. Thanks!

PR #60, opened 2015-11-25
MDLSITE-4323 Add grunt to the prechecker

danpoltawski:
You can use my branch to test with various bits of breakage affected by this patch:
``` https://github.com/danpoltawski/moodle.git grunt-precheck ```

Some notes:
* I didn't end up moving the npm stuff around like we discussed in MDLSITE-4323 - because in the end I didn't actually change very much relating to that.
*  Removing the git reset from grunt_process/grunt_process.sh is possibly controversial - it seemed to me like it should be outside of the grunt scripts purpose to reset the state (i.e. if we need this in the post-checker jobs shouldn't it be done before this script runs?)
* I refactored some of the duplicated checkstyle code into a single php script


danpoltawski:
Note: i've removed the shifter part of this pull request and left it in a branch because we can't merge it till MDL-45364 is fixed really
https://github.com/danpoltawski/moodle-local_ci/compare/grunt-prechecker...danpoltawski:shifter-prechecker

stronk7:
Ok, merged... I'm adjusting the prechecker job in the laptop to use its own npmbase (different from post-jobs). Thanks!

stronk7:
Example run results, with both yui and css not rebuilt (and other errors): http://ci.stronk7.com/view/prehecker/job/Precheck%20remote%20branch/1625/

PR #61, opened 2015-12-03
Use HTTPS for smileys when possible

FMCorz:
Using HTTP breaks the HTTPS validation on the tracker. I removed the pom pom girl as it doesn't have an HTTPS equivalent.

Note: Untested.

### DIFF
--- a/tracker_automations/bulk_precheck_issues/util.sh
+++ b/tracker_automations/bulk_precheck_issues/util.sh
@@ -29,22 +29,20 @@ function negative_tracker_emoticon() {
 
 
 function get_happy_image() {
-    # marys pompoms:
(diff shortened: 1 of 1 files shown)


stronk7:
Pity the pompoms one, it was by far my favorite... we can reintroduce it at any point if we find/serve an https alternative...

stronk7:
Done, thanks!

danpoltawski:
Grr, -1 to this patch! :8ball: 

1) The pompom girl was the best, it's hosted on Mary Evan's site, we can just put it on moodle.org, or wherever with https!
2) Protocol indepdent urls are an anti-pattern: http://www.paulirish.com/2010/the-protocol-relative-url/

danpoltawski:
Fixed in #62

PR #63, opened 2015-12-04
Don't use npm update

danpoltawski:
npm install does the same job and respects the shrinkwrap. Thanks to David for reminding me about this when discussing https://tracker.moodle.org/browse/MDL-52250


stronk7:
Good point, merged! Thanks Dan and David.

PR #65, opened 2015-12-15
MDLSITE-4358 unit tests: move away from template1 as dflt. database

stronk7:
This avoid concurrence problems when 2 jobs are trying to issue
create database statements, rare because it's quick command (1s)
but possible (we have experimented it).


danpoltawski:
Thanks, i guess that this does not affet the public server cos mysql there?

stronk7:
correct, public does everything with mysql.

PR #66, opened 2015-12-19
MDLSITE-4366 savepoints: discard unsafe strings to avoid problems

stronk7:
Now we pre-process the upgrade code, looking for all strings
within it. If the string is considered dangerous (contains
parenthesis and other chars, then it's replaced by safe placeholder
so next regexp operations won't fail).

Safe strings are kept in place in order to keep code transformations
being as small as possible, in case we want to add more checks in
the future (savepoint params checks...).

### DIFF
--- a/check_upgrade_savepoints/check_upgrade_savepoints.php
+++ b/check_upgrade_savepoints/check_upgrade_savepoints.php
@@ -101,6 +101,36 @@ foreach ($files as $file) {
 
     }
 
+    // Extract all string literals in upgrade code, we are not interested on them and can lead to
+    // incorrect calculation of function body later, see MDLSITE-4366. Replace them with simple placeholders.
+    //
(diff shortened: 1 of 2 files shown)


stronk7:
(this also includes minor improvement to shell, to verify all required params are defined)

danpoltawski:
Hehe, I have understand the concept and I spent a bit of time trying to get my head round those regexps but I am still not certain what `(["\'])(?:\\\\\1|.)*?\1'` is doing exactly, but happy to trust you and proceed anyway

stronk7:
one quote (single o double) followed by any character (non-greedy), also including escaped (backslahed) quote found in the first group, ended by the quote found in the first group.

(setting the 2nd group into non-capture mode is just to save some, not needed, memory)

https://regex101.com/r/eR2nV9/1

PR #67, opened 2015-12-22
Check shifter lint errors in the prechecker

danpoltawski:
Can be tested with, for example of problem output:
```git://github.com/danpoltawski/moodle.git shifter-lint-prob```

And when fixed:
```git://github.com/danpoltawski/moodle.git shifter-lint-ok```

### DIFF
--- a/grunt_process/grunt_process.sh
+++ b/grunt_process/grunt_process.sh
@@ -69,7 +69,8 @@ cd ${gitdir}
 rm -fr $(find . -path '*/yui/build' -type d)
 rm -fr $(find . -path '*/amd/build' -type d)
 
-${gitdir}/node_modules/grunt-cli/bin/grunt --no-color | tee "${outputfile}"
+# The echo here works around a problem where shifter is sending colours (MDL-52591).
+echo | ${gitdir}/node_modules/grunt-cli/bin/grunt --no-color | tee "${outputfile}"
 exitstatus=${PIPESTATUS[0]}
 
 # Cleanup symlink as not required after run (and prevent other jobs operating on it)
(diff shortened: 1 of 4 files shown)


danpoltawski:
Note, that the 'error detector' is actually a quite relaxed regexp (which I think is a feature, if you break shifter you'll usually get some prechecker output)

stronk7:
Two questions:

- Those are jshint results from shifter yes? How are they different from own jshint task already in the remote checker?
- Should us be looking for the same in the post-checkers?

And a separate thought:
- Do less errors also come there in error output? Maybe we could also verify them?

danpoltawski:
1. As we agreed in the issue, we should make it the point where we never get shifter lint errors messing up the shifter output (that can be counted as a 'build error'). At the moment, you can get lint errors in shifter which aren't identified by jshint (mostly because the shifter jshint is older) as seen on MDL-45364. Technically I think there is rationale to prevent the jshint task running on yui src folders as [Andrew mentioned](https://tracker.moodle.org/browse/MDL-45364?focusedCommentId=383702&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-383702) - the results from the jshint task are misleading for YUI modules. As I said in above comment, this will actually match other errors (though I think other errors will also cause the js file not to be built which will cause a grunt error on pre-existing files)

2. Yep, makes sense to make this a failure.

3. If there are less errors I think the file will not be built so is caught by the grunt one again. Not sure if we need to do something for it explicitly.

stronk7:
Merged, thanks! Let's follow @ https://tracker.moodle.org/browse/MDLSITE-4329

PR #70, opened 2016-02-03
Change the php code sniffer abbreviation

danpoltawski:
When I see the output of a smurf report with php coding style
problems, I frequently mistake it for the phplint check - which is
much more serious problem.

To avoid this confusion I propose to change the shortname to phpcs to
make it clearer.

An example of me dong this was for [this issue](https://tracker.moodle.org/browse/MDL-39913?focusedCommentId=393168&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-393168).

### DIFF
--- a/remote_branch_checker/lib.php
+++ b/remote_branch_checker/lib.php
@@ -79,7 +79,7 @@ class remote_branch_reporter {
         // Process the cs output, weighting errors with 5 and warnings with 1
         $params = array(
             'title' => 'PHP coding style problems',
-            'abbr' => 'php',
+            'abbr' => 'phpcs',
             'description' => 'This section shows the coding style problems detected in the code by phpcs',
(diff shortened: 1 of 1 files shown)


stronk7:
Perfect. np here. Just pinging @mudrd8mz in case he uses those ids for something in his transformations/processes. Will wait a bit...

mudrd8mz:
Unfortunately my team here has not managed to deliver any automation on fetching and post-processing the smurf results yet. So no, we don't use these ids. You are cleared to merge from us. Thanks for the ping!

stronk7:
Merged, thanks!

danpoltawski:
Turns out this did break `mdk precheck` - see https://github.com/FMCorz/mdk/pull/124

PR #71, opened 2016-02-11
HQ-691 security: JS and CSS compilation conflicts autoresolver

dmonllao:
Tests in https://gist.github.com/dmonllao/6b9720690d1b23fb5081

### DIFF
--- a/rebase_security/rebase_security.sh
+++ b/rebase_security/rebase_security.sh
@@ -1,11 +1,23 @@
 #!/bin/bash
-# $WORKSPACE: place for action to take place (will be used as git repo)
 # $gitcmd: Path to git executable.
+# $gitdir: Directory containing git repo
+# $gitbranch: Branch name
+# $npmcmd: Path to the npm executable (global)
+# $npmbase: Base directory where we'll store multiple npm packages versions (subdirectories per branch)
 # $integrationremote: Remote where integration is being fetched from
 # $securityremote: Remote repo where security branches are being pushed to
-# $branch: Remote branch we are going to check.
 
 set -e
+
+# Calculate some variables.
+mydir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# Ensure (gruntcmd or (recesscmd and shiftercmd)) are available (depends on $gitbranch)
+. ${mydir}/../prepare_npm_stuff/prepare_npm_stuff.sh
+
+moodlecssfile="theme/bootstrapbase/style/moodle.css"
+editorcssfile="theme/bootstrapbase/style/editor.css"
+jsbuild="/yui/build/"
 
 function exit_with_error() {
     echo "ERROR: $1"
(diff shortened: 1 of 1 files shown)


danpoltawski:
(Eloy has suggested this abstraction which is why I think of it) I think that we need to add to this by abstracting the 'grunt setup' part of grunt_process/grunt_process.sh so that then we can ensure grunt is setup on every run.

danpoltawski:
A bit like we have a ``prepare_composer_stuff`` part, we need a ``prepare_grunt_stuff`` which will be used by both jobs

dmonllao:
I've updated the pull request (https://github.com/moodlehq/moodle-local_ci/pull/71/commits/2da5b4234f3dd9bc618ca4248ea8379d96a3de6d) and the tester script https://gist.github.com/dmonllao/6b9720690d1b23fb5081

It is ready to be merged, it is using prepare_npm_stuff, the only thing to comment about is that I've replaced $WORKSPACE and $branch for $gitdir and $gitbranch as required by prepare_npm_stuff.

rebase_security/rebase_security.sh
```diff
@@ -1,12 +1,24 @@
 #!/bin/bash
-# $WORKSPACE: place for action to take place (will be used as git repo)
 # $gitcmd: Path to git executable.
+# $gitdir: Directory containing git repo
+# $gitbranch: Branch we are going to install the DB
```
danpoltawski: There is no installing of DB

dmonllao:
prepare_npm_stuff needs to know the branch, is that a problem or you just mean that I should update the  comment?

danpoltawski:
The reason we use `$WORKSPACE`  is because its the internal jenkins variable for the job directory. Inconsistent variable naming is a problem which seems to be plaguing us everywhere in local_ci.

I much prefer having it better named like this, but it will require us to update all the jobs, an alternative approach is just to setup the variables [like this in remote_branch_checker.sh](https://github.com/moodlehq/moodle-local_ci/blob/09fff4650aa3174bcf58b8f2227c50f53cee50c1/remote_branch_checker/remote_branch_checker.sh#L272).

But I'm not sure, I kinda feel like we should have another issue for rationalising variable names in all our scripts to prevent stuff like this.

danpoltawski:
> prepare_npm_stuff needs to know the branch, is that a problem or you just mean that I should update the comment?

Just update the comment because its a bit misleading (suggesting we are installing a database)

danpoltawski:
> But I'm not sure, I kinda feel like we should have another issue for rationalising variable names in all our scripts to prevent stuff like this.

I talked myself out of that.. if we do it here, its an incremental step towards it (and only breaks one set of jobs at a time).

dmonllao:
Just updated the branch updating the comment and removing a $WORKSPACE reference https://github.com/moodlehq/moodle-local_ci/pull/71/commits/659f42804db58f2016e016c7892ad5b5cdb7677b

danpoltawski:
Thanks David, merged - now reconfiguring all the jobs..

PR #75, opened 2016-02-21
NOBUG: Centralize all the npm install stuff

stronk7:
Right now there are a bunch of jobs requiring the
installation of different nodejs stuff with npm.

A) Chain of jobs by branch.

- For branches < 29 (without package.json file):
  - The job 'shifter_walk' uses 'shifterbase' to maintain different
    shifter versions per branch. (current value ~/shifter_base)
  - The job 'less_checker' uses 'recessbase' to maintain different
    recess versions per branch. (current value ~/recess_base)
- For branches >= 29 (with package.json file):
  - The job 'grunt_process' uses 'npmbase' to maintain different stuff
    (grunt, shifter, less, jshint...) per branch. (current value
    ~npm_base).

B) Prechecker.

  - The job 'remote_branch_checker'' uses 'npmbase' to maintain different
    stuff (grunt, shifter, less, jshint...) per branch. (current value
    ~npm_base_prechecker). Run by 'grunt_process' above.

C) New uses could also require the installation of npm stuff. For sure
   in a different base from the already used above.

So, this new job is a "universal" nodejs stuff installer, behaving as
follows:

1) npm will be in charge of installing any package.
2) every use above (A, B, C) will use a different 'npmbase'.
3) within every 'npmbase' stuff will be installed per branch, aka:
  - A's own npmbase/branchname/node_modules
  - B's own npmbase/branchname/node_modules
  - ....

4) If package.json is available, both grunt-cli and package.json stuff
   will be installed. This is the way in 29 and up.
5) If package.json is not available, but shifterversion has been
   specified, shifter@version will be installed. This is for < 29.
5) If package.json is not available, but recessversion has been
   specified, recess@version will be installed. This is for < 29.
6) At the end of the process all the installed binaries @
   node_modules/bin will be listed.
7) Link the correct node_modules directory in the gitdir being used. The
   job using the installer will be in charge, if needed, of deleting the
   link once the job is done.

Important note: This job does not perform any git operation (fetch,
checkout...). Just requires the gitdir and gitbranch params in order to
decide where packages will be installed and linked to.

Once this job is available, the idea is to use it for new uses (C above)
and also, progresively, replace all the current A & B uses above.

That's it. Next step, replace, in the A, B jobs above, current partial
installers of nodejs stuff by this universal alternative.

### DIFF
--- a/prepare_npm_stuff/prepare_npm_stuff.sh
+++ b/prepare_npm_stuff/prepare_npm_stuff.sh
@@ -68,16 +68,10 @@ if [[ -f ${gitdir}/package.json ]]; then
 
     # Always run npm install to keep our npm packages correct
     ${npmcmd} --no-color install
-    exitstatus=${PIPESTATUS[0]}
-    if [[ ${exitstatus} -ne 0 ]]; then
-        echo "ERROR: Problems installing npm stuff"
-        exit $exitstatus
-    fi
-fi
+else
 
-# Install shifter version if there is not package.json and shifter version has been specified
-# (this is required for branches < 29_STABLE)
-if [[ ! -f ${gitdir}/package.json ]] && [[ -n ${shifterversion} ]]; then
+    # Install shifter version if there is not package.json
+    # (this is required for branches < 29_STABLE)
     shifterinstall=""
     shiftercmd="$(${npmcmd} bin)"/shifter
     if [[ ! -f ${shiftercmd} ]]; then
@@ -98,17 +92,10 @@ if [[ ! -f ${gitdir}/package.json ]] && [[ -n ${shifterversion} ]]; then
     if [[ -n ${shifterinstall} ]]; then
         ${npmcmd} --no-color install shifter@${shifterversion}
         echo "INFO: shifter executable (${shifterversion}) installed"
-        exitstatus=${PIPESTATUS[0]}
-        if [[ ${exitstatus} -ne 0 ]]; then
-            echo "ERROR: Problems installing shifter ${shifterversion}"
-            exit $exitstatus
-        fi
     fi
-fi
 
-# Install recess version if there is not package.json and recess version has been specified
-# (this is required for branches < 29_STABLE)
-if [[ ! -f ${gitdir}/package.json ]] && [[ -n ${recessversion} ]]; then
+    # Install recess version if there is not package.json
+    # (this is required for branches < 29_STABLE)
     recessinstall=""
     recesscmd="$(${npmcmd} bin)"/recess
     if [[ ! -f ${recesscmd} ]]; then
@@ -128,11 +115,6 @@ if [[ ! -f ${gitdir}/package.json ]] && [[ -n ${recessversion} ]]; then
     if [[ -n ${recessinstall} ]]; then
         ${npmcmd} --no-color install recess@${recessversion}
         echo "INFO: recess executable (${recessversion}) installed"
-        exitstatus=${PIPESTATUS[0]}
-        if [[ ${exitstatus} -ne 0 ]]; then
-            echo "ERROR: Problems installing recess ${recessversion}"
-            exit $exitstatus
-        fi
     fi
 fi
 
--- a/prepare_npm_stuff/prepare_npm_stuff.sh
+++ b/prepare_npm_stuff/prepare_npm_stuff.sh
@@ -1,0 +1,146 @@
+#!/bin/bash
+# $gitdir: Directory containing git repo
+# $gitbranch: Branch we are going to install the DB
+# $extrapath: Extra paths to be available (global)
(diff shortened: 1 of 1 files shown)


danpoltawski:
While the current patch makes sense and is flexible for different uses, I think that:
* The recess/shifter versions aren't going to change any time soon for those old branches
* If they are being used, they will be always be in the non package.json present branches 
* They will always be used together in the same branch

So I would propose we can simplify this by:
1) Removing ```shifterversion``` and ```recessversion``` from being optional vars (and in the upstream jobs)
2) Hardcoding the versions in this file (if we ever change it, it'll be the same for all branches anyway)
3) Always installing those hardcoded bins if packge.json isn't present

It'll simplify the logic and also the spread of 'options' through the upstream jobs. I know we specify versions for each branch now, but I don't think its necessary. In future we do it in package.json.

stronk7:
As we have agreed, the job will process package.xml if it exists in dirroot and after that will process matching BRANCHNAME.xml if it exists in the job directory. We'll use that to install shifter, recess and, potentially other stuff in the future if needed.

BTW: Hope npm is able to install things incrementally, never tried...

danpoltawski:
.json 😜

stronk7:
heh, after googling for 1h about how to run arbitrary .xml packages I found it myself hahaha.

stronk7:
As commented in chat, it seems we cannot invoke other package.json files, it's harcoded for a reason.

Doing things improperly I've tried this and has worked, so far:

```
if [ -f package.json ]
    npm install
    mv package.json package.json.saved
fi

if [ -f MOODLE_27_STABLE.json ]
    cp MOODLE_27_STABLE.json package.json
    npm update
fi

if [ -f package.json.saved ]
    mv package.json.saved package.json
]
```

But I don't like it specially. Find original solution way less cheating. Of course, right now, it doesn't matter much as far as both json files are exclusive in our branches (or we run core one for 29 and up, or the "branched" one for 27 and 28). But if the idea was to make things easier to install in the future (csslint...) I'm not sure if that hackery is acceptable.

Any position?

stronk7:
Back to your original comment above... maybe it's enough to apply for a default in the script:

````
$shifterversion=${shifterversion:-0.4.6}
$recessversion=${recessversion:-1.1.6}
````

And done? That way we can get rid of the params safely, still will accept them to be specified. And in the future, if we want to add other stuff per CI purposes, not in package.json... we can do exactly like we do with grunt-cli in the script. NP with that.

Just keeping things simple.... and avoid the multi-packages idea that I'm not happy with.

stronk7:
Ok, I've amended the version a bit, introducing the version defaults and also some github comments along the code. Tell me if that's ok, or I should be using/swtiching to the multi-package alternative or any other idea.

danpoltawski:
The reason I said this originally was just to simplify the script into only two branches 
a) install npm 
b) install shifter/recess

danpoltawski:
I implemented some of my comments.. see https://github.com/moodlehq/moodle-local_ci/compare/laptop...danpoltawski:danp-butchers-eloys-npm-branch

(Sorry, its a bit annoying to do the commits yourself.. I was trying something more radical)

stronk7:
Ok, picking your commit and getting rid of extrapath. I'm happy.

And no, it does not list all those binaries. Those are there coz you've installed them. Only grunt (cli) and shifter are usually there (and recess if < 29).

Oki?

danpoltawski:
In my testing it did list them all

stronk7:
I'd say your npm is really promiscuous, LOL. I've tried here all branches against repo without node_modules (CI situation) and only get grunt&shifter binaries (29 and up). I get, for example, `cake` installed under `./grunt/node_modules/.bin/cake` but not in main one as far as they are not 1st level citizens.

stronk7:
In any case, it's not really important. It's just ainformative listing in case anybody wants, programmatically, verify the availability of some binary. I don't mind if something causes everything to be installed @ first level or no. The output is there, can be sent to null, ignored or processed. Who cares.

stronk7:
Done. I've incorporated your commit and added one moe getting rid of extrapath.

danpoltawski:
Merged - thanks! I'm a bit concerned as to whether the extrapath is doing something i've missed. I suppose we'll see.

stronk7:
I think it was originally needed because the script I copied from had some harcoded `composer` or `npm` executable (or maybe some other). And for those situations, if there is not a global `$xxxxcmd` defined is when the extrapath is useful.

But I remember I replaced a bunch of `npm`cases to the global, forgetting the extrapath in the middle. I bet we'll be able to clean more and more of them. Really think the cmd way is better.

So I don't expect any problem in this script. We'll know soon, once I change grunt, shifter, recess jobs to start using it.

stronk7:
Thanks for merging!

PR #76, opened 2016-04-07
MDLSITE-4419: legacy recess 1.1.9 is the only one for old branches

stronk7:
Using any other (older) version leads to incompatible lessc > 1.4
installed and less files not being processed. So, for the lifes of
27 and 28, we'll be using that 1.1.9 version.

29 and up, not affected, they don't use recess anymore.


danpoltawski:
Merged, thanks

PR #79, opened 2016-06-03
MDLSITE-2839: Check fix version against integration branches

dmonllao:


tracker_automations/check_marked_as_integrated/check_marked_as_integrated.sh
```diff
@@ -0,0 +1,126 @@
...
+# Calculate some variables
+mydir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+basereq="${jiraclicmd} --server ${jiraserver} --user ${jirauser} --password ${jirapass}"
+
+# Include some utility functions
+. "${mydir}/util.sh"
+
+# Let's search all the issues having the "ci" label that have been sent back to development (form peer review or integration)
+${basereq} --action getIssueList \
+           --search "project = 'Moodle' \
```
dmonllao: Yeah, makes sense, agree, will update the patch during next week. This same script could be responsible of adding the label to mark them as green, although we will automatically recheck the issue if there are changes in fixVersions, we will need to remember removing ci-version-check-skip.

danpoltawski: I don't think we need to use the script to do it. This is only in the post-integration step isn't it? 

So, in the case where its expected that the issue should be 'failing' this check, we add the label and done? The check is skipped for that issue forever. (Better than anything more complex )

stronk7:
I think that we only have to mark issues that are violating the expected fixed versions but, still, correct. Than happens, basically, when something goes to 3.0.x (stable) for example, but not not master (some rare regression only in stables). Other than that should not happen that I can imagine.

So I'm with Dan that all we need is to mark ONLY those issues (violating the expectations but correct):

``
humans_are_cleverer_than_bots_and_know_this_is_ok_fixed_versions_wise
``

Slightly offtopic, let me say that, for some reason, I see this more like a hidden comment than a label.

Ciao :-)

danpoltawski:
> Slightly offtopic, let me say that, for some reason, I see this more like a hidden comment than a label.

For once, I might be with you on this suggestion of a comment (because its sort of  an acknowledgement), however I went off the idea because its only possible for admins/original author to reverse the magic comment if someone writes it. (Say an issue gets reopened, and ends with a fix for all versions).

Note that I think there is another related issue to this which could come in future - to detect missing branches on pre-checked issues depending on whether a bug/improvement.

tracker_automations/check_marked_as_integrated/check_marked_as_integrated.sh
dmonllao: ok ok, I get it now. In any case I think it is worth waiting for the script to report a failure to add the label or comment to the issue, so the integrator will still check that the reported missing branches are the expected ones, otherwise we may still miss stuff.

dmonllao:
I've updated the patch to check https://github.com/moodlehq/moodle-local_ci/pull/79/commits/5fd8df45269cd318e12dc753ba734bf488d053a3#diff-cd2a84caa32d025ed27e32e4ff3b48e7R53 I gone with the label because otherwise we should include issues without XXX comment, but I didn't find a way to do it with ~= as other messages match the query (https://answers.atlassian.com/questions/197095/advanced-query-contains-not-and-comments)

danpoltawski:
Merges, thanks

dmonllao:
hehe, I see that you also created jenkins jobs to test it, sorry there is a difference with the format my JIRA-cli (3.9.0) is returning and the one in CI. I've updated the patch to make the grep that extracts the fixed versions more flexible. Can you please take a look? I've updated the branch on top of moodlehq/moodle-local_ci master.

PR #80, opened 2016-06-08
MDLSITE-2839 Expanding regex coverage

dmonllao:
JIRA cli 3.9 and 5.0 seem to return different fixVersion formats,
the regex should be adapted to work with both.

Other changes added as part of this commit:
- Info about skip-ci-version-check label on error messages
- Criteria changed, the issue should be already integrated
- Only the first error was printed


dmonllao:
#79 continuation...

PR #81, opened 2016-06-08
MDLSITE-4620 detect commits which aren't in 'current integration'

danpoltawski:
I guess you can test this by taking an issue 'out' of integration temporarily.

At the moment I don't have a way to make this 'green' if it happens and is a false positive, but maybe thats not so bad.


stronk7:
Thanks Dan, I had missed this completely, sorry.

It looks perfect, looking for all the changes between moodle.git and integration.git and verifying that all them correspond (have in the commit message) to issues actually under current integration.

I can imagine an exception (a couple to be honest):

- NOBUG commits (bumping a version, fixing permissions, svg fixer...)
- Release commits (weekly, Moodle .... release X.x)

In the mean time, i'm going to give it some runs here...

danpoltawski:
I have added af05d15 to address that - although right now struggling to get my bloody jira thing to check it

PR #82, opened 2016-06-08
MDLSITE-2839 relax commit message detector

danpoltawski:
We have the pre-checker to check for messages in the exact right format,
here we should just search for related commits less striclty


danpoltawski:
@dmonllao what do you think?

stronk7:
:+1: (will help epic references and friends)

dmonllao:
I added the ^ to avoid references to other issues but yeah fine for me and for epics although I thought that we were interested in treat these cases manually.

danpoltawski:
It occurred to me this morning that we should probably check for a non-number at the end of the issue id - to prevent MDL-1234 matching MDL-12345

PR #83, opened 2016-06-10
MDLSITE-3782 precheck javascript with ESLint when present

danpoltawski:
ESLint superseeds jshint, when MDL-52127 lands we'll have much better JS coding style checking. This adds patch support to the prechecker whilst providing BC with old branches with jshint. 

After we've given this a bit of time to settle I actually would propose we jshint from the prechecker completely, and make ESlint work on older branches instead (we'd still get lint jshint reported by grunt  for amd/yui files). See also MDL-54889.

### DIFF
--- a/remote_branch_checker/remote_branch_checker.sh
+++ b/remote_branch_checker/remote_branch_checker.sh
@@ -265,9 +265,11 @@ $( grep '<file name=' ${WORKSPACE}/work/patchset.xml | \
 sed '/^$/d' ${WORKSPACE}/work/patchset.files > ${WORKSPACE}/work/patchset.files.tmp
 mv ${WORKSPACE}/work/patchset.files.tmp ${WORKSPACE}/work/patchset.files
 
-# Add .jshint & .csslintrc to patchset files to avoid it being deleted for use later..
+# Add linting config files to patchset files to avoid it being deleted for use later..
 echo '.jshint' >> ${WORKSPACE}/work/patchset.files
 echo '.csslintrc' >> ${WORKSPACE}/work/patchset.files
+echo '.eslintrc' >> ${WORKSPACE}/work/patchset.files
+echo '.eslintignore' >> ${WORKSPACE}/work/patchset.files
 
 # List of excluded paths
 export gitdir="${WORKSPACE}"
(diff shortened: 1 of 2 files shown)


danpoltawski:
You can test this now with my branch:
git pull git://github.com/danpoltawski/moodle.git eslint-prechecker

remote_branch_checker/remote_branch_checker.sh
```diff
@@ -405,9 +407,13 @@ ${phpcmd} ${mydir}/../../moodlecheck/cli/moodlecheck.php \
 # Exclude build directories from the results (e.g. lib/yui/build, lib/amd/build/)
 find $WORKSPACE -type d -path \*/build | sed "s|$WORKSPACE/||" > $WORKSPACE/.jshintignore
 
-# Run the JSHINT (using the checked out .jshint file)
-${jshintcmd} --config $WORKSPACE/.jshintrc --exclude-path $WORKSPACE/.jshintignore \
-    --reporter=checkstyle ${WORKSPACE} > "${WORKSPACE}/work/jshint.xml"
+if [ -f $WORKSPACE/.eslintrc ]; then
+    ./node_modules/.bin/eslint -f checkstyle $WORKSPACE > "${WORKSPACE}/work/eslint.xml"
```
danpoltawski: This raw use of node_modules might be controversial, thoughts welcome. (I think we should use the grunt installed version because we then don't need to worry about version dependencies.

stronk7:
As commented in the chat... what if we use ```$(${npmcmd} bin))``` to know where to find the executables?

danpoltawski:
> As commented in the chat... what if we use $(${npmcmd} bin)) to know where to find the executables?

I have updated the patch to do that (and also error more gracefully if the npm file isn't found)

danpoltawski:
Note that I updated the eslint-prechecker branch on top of latest in MDL-52127

stronk7:
Ping, maybe you should move this on top of #85, for your consideration, ciao :-)

danpoltawski:
I have rebased on top of #85 - I changed the approach slightly different to what it was before - instead of the if/else I do the eslint stuff before the node_modules directory is deleted, and then check for the presenece of eslint.xml

stronk7:
Installed in the laptop, running a couple of tests (with and without .eslintrc) ... before merging this into production...

stronk7:
Worked ok against your branch adding .eslintrc, grunt installed it ok and warnings were detected:

http://ci.stronk7.com/view/prehecker/job/Precheck%20remote%20branch/1710/

stronk7:
And also ok against current master, fall-backing to jshint:

http://ci.stronk7.com/view/prehecker/job/Precheck%20remote%20branch/1711/

stronk7:
So, merged! Let's break the unheld issues... thanks!

PR #84, opened 2016-06-10
NOBUG: Detect grunt failures properly

danpoltawski:
I'm slightly confused because i'm sure i've seen this working, but right
now if the grunt command fails it immediately exits without completing
other parts of the script and doens't add an error in the the
grunt-errors.txt. That meant the prechecker didn't detect any failures..

I don't generally like swiddling set -e and set +e, but this was least destructive way

### DIFF
--- a/grunt_process/grunt_process.sh
+++ b/grunt_process/grunt_process.sh
@@ -52,8 +52,8 @@ gruntcmd="$(${npmcmd} bin)"/grunt
 if [ -x $gruntcmd ]; then
     set +e
     $gruntcmd --no-color > >(tee "${outputfile}") 2> >(tee "${outputfile}".stderr >&2)
-    set -e
     exitstatus=$?
+    set -e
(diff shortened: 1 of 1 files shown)


danpoltawski:
I discovered this because a failed grunt build was not generating any prechecker issue

stronk7:
Think this is now part of #85, so think this can be closed once that one is merged.

danpoltawski:
Yep, lets close this to avoid confusion and will test in #85

danpoltawski:
I'm actually gonna reopen this, because I think there is some more work to be done here. We now get failures handled ok-ish, but we should report that grunt failed, not that the files failed to build

danpoltawski:
Self merging this simple fix

danpoltawski:
There is a remaining issue: https://tracker.moodle.org/browse/MDLSITE-4625

PR #90, opened 2016-06-14
HQ-1070 don't use colors in grunt output

danpoltawski:
I am sure grunt is supposed to autodetect this, but it doens't seem to be, so avoid those horrible colour escape sequences.

Seen here:
http://integration.moodle.org/job/99.%20SECURITY.%20Rebase%20security%20branch%20(30_STABLE)/549/console
```
[4mRunning "less:bootstrapbase" (less) task[24m
[32m>> [39m2 stylesheets created.
```


stronk7:
running....!

PR #91, opened 2016-06-15
MDL-52127 eslint: ignorefiles task is not there any more

stronk7:
ust decided to comment the line instead of deleting in case there
is a chance of getting it back when the subdirectories problem
is fixed and the task is back. See MDL-54930 for more info


danpoltawski:
What do you think we should do about this pull request? Technically I don't think we should run `grunt ignorefiles` here because its tracked by git.

danpoltawski:
Re: above comment, I discovered that this was buggy - so I pushed a commit which explained the situation and removed `grunt ignorefiles` completely. If we are going to run that command, we need to add it to a point in the checks where we have the entire moodle checkout available

stronk7:
(y) well done!

PR #93, opened 2016-06-20
MDLSITE-4644 prechecker: make eslint to run out from exit-on-error

stronk7:
It seems that in the last move we did put eslint under an
exit-on-error section within the codechecker. That was leading
to executions exiting with error when eslint detected some
coding-style  problem.

Note the fix is not the perfect fix, just a quick attempt to keep
the things working. There are some linked issues where real
improvements should happen:

- MDLSITE-4625, about to properly inspect exit codes an output
    for every check and be able to discern correct executions
    from incorrect ones.
- MDLSITE-4211, about to have the codechecker, and other local_ci
    stuff covered by own unit tests.


stronk7:
Added new commit to report both the node and npm versions installed in the server.

stronk7:
self-merged, thanks!

PR #95, opened 2016-06-23
MDLSITE-3831 Improve handling of both binary and with-whitespace files

stronk7:
Do not integrate yet, being tested with all sort of patchsets.


stronk7:
Oh, my mistake I did pick the target branch wrongly in this PR, please, move on to #96. Everything is awaiting there.

PR #96, opened 2016-06-23
MDLSITE-3831 Improve handling of both binary and with-whitespace files

stronk7:
This PR contains 2 commits,

- the first one just ensures that paths and files containing whitespaces are properly handled. Note that core does not have files with whitespaces, but still because of another issue, it was reproducible when checking a plugin.
- the second one changes the diff processor and now it's able to handle both binary files and files with whitespaces. Previously they were not being detected... and the prechecker was deleting all them. That was not problematic because those binary files are not subject to any analysis, but was inaccurate and, potentially, prone to cracks. Now t just works and all files are processed.

For more details and tests, https://tracker.moodle.org/browse/MDLSITE-3831 is the issue :-)


stronk7:
Thanks!

PR #97, opened 2016-07-03
MDLSITE-4667 move smurf.html to https friendly stylesheet

danpoltawski:
(Also fixed some docs urls I noticed whilst greping)

We to fix the CSP for this too


danpoltawski:
Self merged because it seemed trivial

stronk7:
🆗

PR #99, opened 2016-07-12
MDLSITE-4678 grunt_process: run grunt ignorefiles and detect changes

danpoltawski:
I actually developed this by added a bats fixture for this on the unit testing branch before developing, have a look:

 https://github.com/moodlehq/moodle-local_ci/commit/218e58069983c26d83ce3958ae0cc049927a23cc 

Was a useful exercise writing the tests first, I discovered:

1. You need to do the grunt part before npm symlink is disabled (eliminating my initial attempts to run grunt twice, and eventually realising to just specify both tasks)
2. Need to remember to only run it on 3.2+

But [bats is still crap on travis](https://travis-ci.org/danpoltawski/moodle-local_ci/builds/144131676), so bah for my test driven development experience being ruined!


stronk7:
Looks perfect, giving it a run...

PR #100, opened 2016-07-13
MDLSITE-4685: fix problems in illegal_whitespace.sh and add tests

danpoltawski:
I started this fix by first adding tests verifying the current behaviour, then fixing the bug


danpoltawski:
In case you didn't see the issue - this fixes the grep errors seen here https://integration.moodle.org/job/02.%20Illegal%20whitespace%20(master)/5980/console

stronk7:
wow, i never saw those warns (or forgot them).

stronk7:
Merged, thanks!

PR #101, opened 2016-07-13
Remove some unused scripts

danpoltawski:
They will still be available in the git history, but by removing them we make it clearer what is an active script. (Partly because I wanted to get a good handle on what scripts need tests adding to them, but I don't think its good practice for us to leave 'unmaintained' scripts lying around).


danpoltawski:
I'm assigning to @stronk7 because he is the best person to determine if my assesment of the files as unused is correct

stronk7:
Uhm... I'm not really sure it's a good idea to remove all, the ram disk ones, for example, are fully working or also the:

- project size report: http://ci.stronk7.com/view/reports/job/R00.%20Project%20size%20report%20(master)/plot/
- generate_component_ant_files: that gives us this http://ci.stronk7.com/view/qa-todo/job/Q06.%20Project%20todo%20tasks%20(master)/160/tasksResult/

others are disabled in http://ci.stronk7.com (copy/paste, phpmd...) and others for sure are deletable (simpletests, configure site)... so it's really heterogeneous stuff.

I'd say I'd keep the 3 commented (ram disk, project size, generate component), I'm ok also if those 3 are moved to "util/support" dir, as you wish. And I'm happy with all the rest being deleted (I will delete the old jobs using them in the laptop CI).

Thoughts...?

danpoltawski:
Sure - I have updated the patch keeping those 3.

Where is the ram one used?

stronk7:
I had it here working for some time (mac) preparing for https://tracker.moodle.org/browse/MDLSITE-2913 to be used everywhere. I bet at some point we'll be revisiting it, that's the reason I think it makes sense to keep them.

stronk7:
Merged thanks!

PR #102, opened 2016-07-14
MDLSITE-4685: fix detect_conflicts.sh and improve testing infrastructure

danpoltawski:
- Introduce tests for detect_conflicts
- Fix spaces problem in detect_conflicts
- Add some setup tests for gnu tools in use (can fail when my mac isn't setup right)
- Introduce git_apply_fixture and tidy up tests to avoid having to keep cd'ing

### DIFF
--- a/detect_conflicts/detect_conflicts.sh
+++ b/detect_conflicts/detect_conflicts.sh
@@ -2,6 +2,16 @@
 # $gitdir: Directory containing git repo
 # $gitbranch: Branch we are going to examine
 
+set -e
+
+required="WORKSPACE gitdir gitbranch"
+for var in ${required}; do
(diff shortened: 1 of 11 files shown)


stronk7:
Sorry, silly question... where does OLDPWD come from? I only can it being used but never set.

stronk7:
Ah! It's built-in bash!

danpoltawski:
ya https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html

stronk7:
I like, this, thanks for modernizing the old script and the new git_apply_fixture(). It makes things easier and mistakes-savvy. yay!

stronk7:
Done, cool!

PR #104, opened 2016-07-20
MDLSITE-4695 prechecker: add ability to push precheck branches to gitlab

danpoltawski:


danpoltawski:
I've chosen not to fix  MDLSITE-3676 at the same time, but it will straight forward to do

stronk7:
And the tests?

stronk7:
Merged, thanks!

danpoltawski:
The tests for pre-checker are going to be the most critical, but quite a bit to do before that :)

PR #106, opened 2016-07-21
Add tests to diff_extract_tests and remove Moodle depency

danpoltawski:
Add tests to diff_extract_tests and remove Moodle depency


stronk7:
Thanks Dan, they look perfect, are passing... so let's push them.

PR #107, opened 2016-07-21
List valid components tests

danpoltawski:
Even though it has an --absolute option you still end up with a raw path, so have to sed the fixture file.


tests/02-list_valid_components.bats
```diff
@@ -0,0 +1,26 @@
...
+
+    OUTPUT=$WORKSPACE/valid_components_out.txt
+
+    # Run test
+    php list_valid_components.php --basedir=$gitdir > $OUTPUT
+    assert_success
+    diff -ruN $EXPECTED $OUTPUT
+    assert_output ''
+
+    rm $EXPECTED $OUTPUT
```
stronk7: Just guessing if:

1) This should be a call to ```clean_workspace_directory()``` instead, or, better
2) Guessing if BATS does have a corresponding teardown() method where we can automate that.

No matter of it, I'm proceeding with this, is perfect.

stronk7:
Thanks!

tests/02-list_valid_components.bats
danpoltawski: @stronk7 
1) Yeah you are probably right. When i first wrote the test I wasn't using the workspace directory
2) [It does](https://github.com/sstephenson/bats#setup-and-teardown-pre--and-post-test-hooks), though I am not sure we want to clean up on every execution. I do clean up the workspace dir [on the start of each test run ](https://github.com/moodlehq/moodle-local_ci/blob/c523b79/tests/libs/shared_setup.bash#L69) - but I did this because I found there were situations when I killed the tests and then the teardown type operations were not run (well, understandably).

PR #109, opened 2016-07-26
Remove Moodle dependency from remote_branch_checker/remote_branch_reporter.php

danpoltawski:
Add tests and remove the Moodle dependency from `remote_branch_checker/remote_branch_reporter.php`.

I think this is the last script before we can start to get `remote_branch_checker.sh` tested :)


stronk7:
I think that absolute paths were used there because initially we were planning to show links to the lines in place... but that never came to reality. Maybe we could consider making them relative to moodleroot and forget. For sure that will lead to cleaner and more transportable results. Thoughts?

stronk7:
(note maybe there is some other dependency of those full paths I don't remember but I bet we'll be able to workaround it).

stronk7:
Merged, thanks!

danpoltawski:
There are some paths we can't get rid of because external tools generate them I think

stronk7:
Oops, hope https://tracker.moodle.org/browse/MDLSITE-4702 is not related with this... investigating...

stronk7:
Still looking why... but with this reverted I get the jobs passing locally, while they fail with it applied.

stronk7:
Wow, found it. Was really well hidden! (db90151)

danpoltawski:
Looking improving the tests to avoid this in https://github.com/moodlehq/moodle-local_ci/pull/110

stronk7:
(thanks, Dan!)

PR #110, opened 2016-07-27
MDLSITE-4702: Convert tests to neutral path to prevent regression

danpoltawski:
Update the tests so that they don't run in the directory of the script to prevent a repeat of [MDLSITE-4702](https://tracker.moodle.org/browse/MDLSITE-4702).

Not ready for merging because i haven't converted  `99-detect_conflicts.bats` or `99-illegal_whitespace.bats` yet - because when I did they break! 

I don't love the variable name test `filedir` - suggestiosn welcome

### DIFF
--- a/detect_conflicts/detect_conflicts.sh
+++ b/detect_conflicts/detect_conflicts.sh
@@ -52,8 +52,8 @@ then
     prevcount=`tail -1 "$countfile" | cut -s -f3`
 fi
 
-# Count and send to countfile
-count=`cat "$lastfile" | wc -l`
+# Count and send to countfile. The tr is to remove leading spaces on some platforms..
+count=`cat "$lastfile" | wc -l | tr -d '[[:space:]]'`
 echo "$BUILD_NUMBER	$BUILD_TIMESTAMP	$count" >> "$countfile"
 
 # Get best count ever or create it
--- a/illegal_whitespace/illegal_whitespace.sh
+++ b/illegal_whitespace/illegal_whitespace.sh
@@ -52,8 +52,8 @@ then
     prevcount=`tail -1 "$countfile" | cut -s -f3`
 fi
 
-# Count and send to countfile
-count=`cat "$lastfile" | wc -l`
+# Count and send to countfile. The tr is to remove leading spaces on some platforms..
+count=`cat "$lastfile" | wc -l | tr -d '[[:space:]]'`
 echo "$BUILD_NUMBER	$BUILD_TIMESTAMP	$count" >> "$countfile"
 
 # Get best count ever or create it
--- a/phplib/clilib.php
+++ b/phplib/clilib.php
@@ -65,7 +65,7 @@ function cli_input($prompt, $default='', array $options=null, $casesensitiveopti
             $input = strtolower($input);
         }
         if (!in_array($input, $options)) {
-            cli_writeln(get_string('cliincorrectvalueretry', 'admin'));
+            cli_writeln('Incorrect value, please retry');
             return cli_input($prompt, $default, $options, $casesensitiveoptions);
         }
     }
--- a/remote_branch_checker/remote_branch_reporter.php
+++ b/remote_branch_checker/remote_branch_reporter.php
@@ -54,7 +54,7 @@ list($options, $unrecognized) = cli_get_params(array(
 
 if ($unrecognized) {
     $unrecognized = implode("\n  ", $unrecognized);
-    cli_error(get_string('cliunknowoption', 'admin', $unrecognized));
+    cli_error("Unrecognised options:\n{$unrecognized}\n Please use --help option.");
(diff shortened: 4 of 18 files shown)


stronk7:
would be maybe, better to have a custom "run" command, doing the thing under the hood and you just have to make:

````
our_own_run relative/path/to/whatever/we/want/to/execute
````

keeping that $filedir calculations and uses well hidden?

alternatively a new $BASE_PATH, could be used and then using

````
run $BASE_PATH/relative/path/to/whatever/we/want/to/execute
````

BASE_PATH or BASE_SRC or BASE or DIRROOT or...

Just sharing as I imagine the alternatives.

**Comment by danpoltawski** [comment]
Yes, great idea. I've updated the patch - included the broken tests.. so expect travis to fail.

**Comment by stronk7** [comment]
detect_conflicts failing?

**Comment by stronk7** [comment]
And... why is travis passing? because previous job was passing or what?

**Comment by danpoltawski** [comment]
> And... why is travis passing? because previous job was passing or what?

Turned out my issue was a portability issue with bsd wc -l adding whitespace. I added 8d215c7 to help (could've made the tests more tolerant, but makes more sense to get consistent output)

**Comment by danpoltawski** [comment]
For the record, if you revert db90151 you'll see the tests fail as expected.

**Comment by stronk7** [comment]
testing locally... with and without db90151 ...

**Comment by stronk7** [comment]
Uhm, I'm getting here some failures with whitespace checks:

````
✗ illegal_whitespace: first run OK
   (from function `assert_output' in file tests/libs/bats-assert/src/assert.bash, line 231,
    in test file tests/99-illegal_whitespace.bats, line 31)
     `assert_output --partial "current count: 959"' failed
   
   -- output does not contain substring --
   substring (1 lines):
     current count: 959
   output (6 lines):
     HEAD is now at 47806ea Moodle release 3.1.1
      current count: 1224
     previous count: 999999
         best count: 999999
     same/better results than previous counter
     got best results ever, yay!
   --
````

I ran it twice, the second one cleaning completely the git clone just in case there was there something unexpected. Trying again with a brand new clone...

**Comment by stronk7** [comment]
Same with brand-new clone. I continue getting 1224.

**Comment by danpoltawski** [comment]
Do you get the same before the patch? And have the correct grep?

**Comment by danpoltawski** [comment]
Here are my results of the workspace directory after my run: 
[workspace.zip](https://github.com/moodlehq/moodle-local_ci/files/386989/workspace.zip)

Could you try applying this patch and comparing the results:
```
diff --git a/tests/99-illegal_whitespace.bats b/tests/99-illegal_whitespace.bats
index ca53191..41b1223 100755
--- a/tests/99-illegal_whitespace.bats
+++ b/tests/99-illegal_whitespace.bats
@@ -19,6 +19,8 @@ teardown () {
     cp -R $WORKSPACE/. $statedir
 }
 
+echo $WORKSPACE >&2
+
 @test "illegal_whitespace: first run OK" {
     # Ensure initial state is clean.
     clean_workspace_directory
@@ -36,6 +38,8 @@ teardown () {
 }
 
 @test "illegal_whitespace: normal state OK" {
+    skip
+
     # On second run, should still pass with same results
     ci_run illegal_whitespace/illegal_whitespace.sh
 
@@ -47,6 +51,7 @@ teardown () {
 }
 
 @test "illegal_whitespace: failure reported when whitespace error detected" {
+    skip
     # Lets introduce a whitespace error and ensure it fails
     git_apply_fixture 31-whitespace-error.patch
 
@@ -59,6 +64,7 @@ teardown () {
 }
 
 @test "illegal_whitespace: clean up state" {
+    skip
     # Not a real test, just allows us to avoid storing state after
     # the rest of the test suite has run.
     clean_workspace_directory
```

stronk7:
Bah, I've executed them again twice, in the very same terminal session than previous ones, without having modified anything in the middle and now they are passing completely.

````
 ✓ illegal_whitespace: first run OK
 ✓ illegal_whitespace: normal state OK
 ✓ illegal_whitespace: failure reported when whitespace error detected
 ✓ illegal_whitespace: clean up state
````

how is that possible... looking history....

stronk7:
Ohoh....

before:
´´´´
LOCAL_CI_TESTS_GITDIR=/Users/stronk7/git_moodle/ci-tests/
´´´´
after:
´´´´
LOCAL_CI_TESTS_GITDIR=/Users/stronk7/git_moodle/ci-tests
´´´´

Confirmed, just added the trailing slash back and got the failures again.

danpoltawski:
I can't reproduce that - can you see what the differences in output are?

stronk7:
Ok, I've confirmed that I also get the same failures without your patch... so merging it and moving to separate issue the trailing slash problem (sad you cannot reproduce it).

stronk7:
Merged, thanks!

PR #111, opened 2016-07-29
slashes, define_excluded and tests

stronk7:
This solves the probem detected in #110 when trailing slashes were being passed.

Also introduces a new option (format) to get the script outputting results instead of sharing variables (useful for tests for now, but can imagine us moving all current uses to that).

And, finally, tests covering the output of all "formats" and also the slashes problem.


stronk7:
The only problem I'm facing is that, for some reason, linux sorting does weird things with asterisks, I've tried disabling globbing (set -f), enforcing C lang/lctype... no luck. So sharing this in advance in case anybody knows an easy solution. Need results sorted to be able to match them.

define_excluded/define_excluded.sh
```diff
@@ -99,7 +105,7 @@ if [[ -n ${gitdir} ]]; then
 fi
 
 # Sort and get rid of dupes, they (maybe) are legion.
-excluded=$(echo "${excluded}" | sort -u)
+excluded=$(LANG=C; echo "${excluded}" | sort -u)
```
stronk7: This is the problematic line, sorting asterisks, dots... in a non ascii way.

stronk7:
Yay, LC_ALL=C did the trick, apparently, now getting portable sorted results.

PR #112, opened 2016-07-29
No bin bash anymore

stronk7:
That way the bash used is the specified one (on every case PATH precedence) and not the
default one installed in the system.

Covered with a test looking for old uses.


danpoltawski:
Thanks!

PR #113, opened 2016-07-29
Verify commit messages tests

danpoltawski:
I spent a lot longer on this than I wanted to due to stupid bash!


stronk7:
Perfect, thanks!

PR #114, opened 2016-07-30
Verify more GNU coreutils depenencies and reverse course with wc tolerance

danpoltawski:
This reverses course (and mostly reverses) changes in 89a64f3d214511 and b76bc6f07 to make the scripts more tolerance to mac/bsd wc.

After I started looking at versions_check_set I found another gnu depdency and I think its best to make the scripts simpler and less portable. If there is one or two gnu dependencies, its not worth having tolerances for other things like wc.

Also make verify_commit_messages/verify_commit_messages.sh have +x permission.


stronk7:
Perfect, thanks!

PR #116, opened 2016-07-30
Add tests for check_upgrade_savepoints

danpoltawski:


stronk7:
Seems that we are facing some hash length problems:

https://travis-ci.org/stronk7/moodle-local_ci/builds/148650809

danpoltawski:
I added https://tracker.moodle.org/browse/MDLSITE-4704 to track that bug

PR #117, opened 2016-07-31
NOBUG: Continuously verify perms are ok

stronk7:
While reviewing #114 saw you changed an incorrect perm and then performed some searches here. Then I thought maybe it was not bad to move those searches to be continuously performed, so I moved them to a new test.

Note: The only point I'm not 100% sure are the ````".bash"```` extension files. I've added them to the ````$nonexecutables```` list of extensions (to make the match current status), but IMO they should be changed to ````755````and moved to the executables list (or maybe renamed to ````.sh```` although some of them, upstream, are already ````.bash````.

Thoughts?


tests/00-verifications.bats
```diff
@@ -8,3 +8,57 @@ load libs/shared_setup
...
+# This test verifies permissions of directories and important files are correct.
+@test "repository file & dir permissions are correct" {
+    # Define executable and non-executable extensions
+    executables=('sh' 'bats')
+    nonexecutables=('bash' 'html' 'jar' 'md' 'out' 'patch' 'php' 'template' 'txt' 'xml' 'xsl' 'yml')
+    # Other nonexecutable files, but looked by name, not by extension
+    otherfiles=('.gitignore' '.htaccess')
+
+    # Directories must be 755
+    run find $BATS_TEST_DIRNAME/.. -type d -not -perm 755 -not -path "*/.git/*"
```
stronk7: Pity, for some reason this is failing in travis, looking for dirs not having 755. Passes here. Will try some alternative solution.

stronk7:
Aha, it does not have mush sense to check directories, I bet. Can see 2 alternatives:

1. Or we ````umask 022```` so directories are created (git checkout/clone) ````755````
2. Or we completely get rid of the directories test, as far as we are really testing nothing.

Think I'm going to go towards the 2nd alternative. Commenting it...

stronk7:
Done, dir permission tests commented out.

stronk7:
Pity, also failing while testing file perms, sniff, sniff.

stronk7:
Ok, one more try, now checking only for the executable bits for ````uga````. It seems that git only traces those (I've re-added the folders, just to see how they behave with travis).

stronk7:
Working. Final version with folder perm checks commented (git does not manage those at all). Only execution bits of files

stronk7:
And this just shows how an incorrect change (a095acecd3726c5a4dd5b3eed21ed22ede8b63e8) leads to tests failure:

https://travis-ci.org/stronk7/moodle-local_ci/builds/148690844

So, the only remaining point is if we should consider .bash files in the executable group and set the bit for them (right now they are in the non-executable group).

Ciao :-)

danpoltawski:
Technically I think we should do `.bash` too given the only reason I didn't call that shared file `.sh` is because the bats load thing requires it to be named .bash. But we can do that in a new commits.

Also, despite me changing`.bats` to be executable - I don't actually think they need to be that - I mostly just wanted my terminal colours to be that, i've been running them through the bats command the whole time. I think I saw that you were running bats locally from a git dir too - so maybe we should enforce them to be not executable.

danpoltawski:
Given my comments why did I merge ;-) I thought i'd merge this change to keep things under control and we can do any blanket changes in a new issue..

stronk7:
Ok, so the idea is:

- make current .bash executable .
- make .current bats non-executabel.
- configure the test accordingly (moving those extensions to their new statuses).

stronk7:
https://tracker.moodle.org/browse/MDLSITE-4705 created for these remaining points...

PR #118, opened 2016-08-02
Simplify prechecker a bit

danpoltawski:
Don't create a local (e.g. in case of master) 'master' branch, only the 'master_precheck'
branch is needed locally so don't do work to create the 'master' one.

We can also use git checkout -B to simplify (as used in security checker).

This is to prepare for testability..


danpoltawski:
The reason for this is because I think to test the prechecker we're gonna have to add some sort of special 'reset' stage to set itself to a constant base. So I am trying to simplify that here.

danpoltawski:
Closing in preference of #119

PR #119, opened 2016-08-03
Preparation for prechecker testability

danpoltawski:
I think that we are going to need to 'reset' the base at which we merge at - so that we can set the prechecker running _a point in time_.

So these commits are to prepare for that.

Best to review each commit one at a time.


stronk7:
done, thanks!

PR #120, opened 2016-08-03
Add tests for prechecker

danpoltawski:
As part of this:

- Introduced assert_files_same and changed existing tests to use it
- Skipped local_moodlecheck in the testing environment. This is a bit hacky, but better than any more extensive changes right now IMO
- Removed the custom phpcs runner and replaced it with the official phar runner (the existing runner wanted a local_codechecker path and didn't seem worth passing along) 
- Allowed the coding style dir to be passed as an argument - slightly related to the last point. We could've passed the whole local_codechecker path - but I do wonder about us future moving that cs path into moodle core (blue sky)
- Allow the branch point to be reset (as supported by the previous commits to prechecker)
- Raised the node version in travis (required for eslint to succesfully run)


danpoltawski:
Note that the current 'old branch failing' test we probably should get rid of when we next add a fixture - I just added it to prove that we can make a really old branch work - and it won't break randomly in the future.

stronk7:
The only point I'm not sure here is about the new handling of the phar/codechecker... 

1) We should at very least print the version of phpcs bundled or keep it documented (readme...).

2) it's not clear for me, in fact, if we should be bundling it or, instead, start using composer to install and update it. Can imagine a point were the standard itself will be separate package, easy to get also via composer and use the new parameter, cleaning any dependency from local_codechecker.

3) why local_codechecker is installed from within setup() and not .travis ? Same point than above, once we move to composer... it easy to delete the dependency and switch to the new one without many changes in tests, just extra param and done ````phpcsstandard=vendor/mooodlehq/phpcs-standard/moodle````(for example).

Other that those details to agree on... beautiful PR !

danpoltawski:
Re: 1/2) - so do you propose we add a composer.json and make sure we run composer as part of the install/deployment of it?

3) The main reason I did that was to avoid the 'dependencies' for running the tests - I was thinkiing it would be a small pain to introduce another LOCAL_CI var to run the tests

danpoltawski:
I've pushed two commits..

- Moving local_codechecker out of the tests
- Switching to composer for phpcs

Note that I did the bare minimum to make the permissions tests pass - I tried to make the find ignore paths more configurable because we should exclude vendor from all of them, but I was getting annoying problems so switched to the minimum approach.

stronk7:
Have added this to the [autoupgrade job](http://ci.stronk7.com/view/various/job/Moodle%20CI%20Auto%20Upgrade/), both in laptop and public CI servers:

````
# Ensure composer stuff is installed/updated on local_ci changes
cd ${WORKSPACE}/local/ci && ${composercmd} install
````

And composer stuff was installed ok:

http://ci.stronk7.com/view/various/job/Moodle%20CI%20Auto%20Upgrade/572/

And a prechecker job executed on top of that, too (has phpcs warnings):

http://ci.stronk7.com/job/Precheck%20remote%20branch/1731/

So, apparently, all, ok, going to merge this so public server will start using it too. Thanks!

stronk7:
Yay!

PR #121, opened 2016-08-03
NOBUG: Make sure clean_worspace_directory removes .files

danpoltawski:
My original implementation of this was to prevent accidents with unset
variables causing an rm -rf /.

But I don't see a neat way to do that..


stronk7:
Thanks!

PR #122, opened 2016-08-03
NOBUG: Clean up the code for storing $WORKSPACE between tests

danpoltawski:
* Centralise the code and add tests
* Don't store it in the cache directory - store it in a tmp directory
  which only persists per file/run
* Add helpers for determining when in the first or last test


stronk7:
Clever one! Thanks!

PR #123, opened 2016-08-04
Utilise stylelint in the prechecker (with tests)

danpoltawski:
Based on top of #120 

Taken the same approach as jshint/eslint (although I think particularly with csslint we should get rid of it soon, its doing so little).

Haven't added tests for the reporting for remote_branch_reporter.php yet


stronk7:
Sweetly sweet! Thanks!

PR #126, opened 2016-08-11
Ensure stylelint returns a prechecker result, even if not files to check

danpoltawski:
Fixes a bug which Fred reported - where the css part of report isn't returned when stylelint runs.

I've added a fixture for the branch which I saw this behaviour on


danpoltawski:
Self merging, but at least we have tests ;-)

danpoltawski:
Here is it working https://integration.moodle.org/job/Precheck%20remote%20branch/26297/console

PR #125, opened 2016-08-09
MDL-55465 Respect .stylelintignore

danpoltawski:
Shouldn't be merged until this weeks weekly is published (including MDL-55465)

PR #127, opened 2016-08-16
Replace unreliable fixture

danpoltawski:
The fixture introduced in #126 seems to have unreliable results.. which
caused variable phpcs results in #125 - the failures seemed unrelated
and not part of what was being tested, so simplify that fixture to stop
me going crazy.. (hopefully)


danpoltawski:
I don't love doing this without working out the problem.. but look at this build failure:
https://travis-ci.org/moodlehq/moodle-local_ci/builds/151829803

The change is:
https://github.com/moodlehq/moodle-local_ci/commit/a768658ebcef5b472dec485362964d91b20c428a

Ignore the deletions (they were just to try and exclude stuff).. the 'failure' is caused by a new commented out line.. aka no changes to the code.

PR #129, opened 2016-08-19
MDLSITE-4440 verify_commits: check amos commands

danpoltawski:
Uses code from AMOS (but copy and pasted).

### DIFF
--- a/tests/02-checkstyle_manipulations.bats
+++ b/tests/02-checkstyle_manipulations.bats
@@ -25,6 +25,10 @@ assert_checkstyle() {
     assert_checkstyle verify_commit_messages/commits2checkstyle.php commits.txt commits.xml
 }
 
+@test "verify_commit_messages/commits2checkstyle.php info only " {
+    assert_checkstyle verify_commit_messages/commits2checkstyle.php commits-info.txt commits-info.xml
+}
+
 @test "remote_branch_checker/checkstyle_converter.php: phplint" {
     assert_checkstyle 'remote_branch_checker/checkstyle_converter.php --format=phplint' phplint.txt phplint.xml
 }
--- a/tests/02-verify_commit_messages.bats
+++ b/tests/02-verify_commit_messages.bats
@@ -90,7 +90,7 @@ commit_apply_fixture_and_run() {
 @test "verify_commit_messages/verify_commit_messages.sh: AMOS incomplete commands" {
     commit_apply_fixture_and_run amos-incomplete.patch
     assert_failure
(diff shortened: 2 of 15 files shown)


stronk7:
I like this, but the copy/paste code part... just guessing if it wouldn't be better to:

1) have AMOS clone at hand.
2) extend the class using some late binding if needed in base.

That way we'd react immediately to any change in AMOS... opinions?

danpoltawski:
I feel like its overkill - I especially feel like doing it that way would lead it to become potentially fragile for unrelated changes which are going on in amos (which might increase the requirements of the script and so on). I doubt those parsing functions have ever changed in recent years in any case - and I imagine we would be adding more functionality to detect non-matching amos script (I mean I imagine the prechecker would get more parsing powers than amos would).

Note: I would've just copy and pasted the regexps, but I did it this way because I thought there was less change of me getting things wrong and it looked relatively clean to do and also provide some nice informative messages about what would happen. I think i'd rather just switch to the basic regexps instead of pulling AMOS in as a dependency.

Pinging @mudrd8mz

mudrd8mz:
I agree with Dan. Pulling AMOS as a dependency sounds like an overkill. I am pretty sure these regex's won't change in years, if ever. We will sooner have troubles with UNIX timestamps.

But, I realised another important thing. I did not study the attached unit tests well. But please note that AMOS only processes commits that somehow modify some */lang/en/*.php file. So if there is a commit with AMOScript that modifies README.txt only, AMOS would not parse it. If possible, that would be a good precheck, too.

mudrd8mz:
https://github.com/moodlehq/moodle-local_amos/blob/master/cli/parse-core.php#L367-L398

stronk7:
Thanks, I'm happy then.

About to verify that the commit being analyzed does include some lang/en/xxx file, sounds like we could do that by quickly looking/grepping to the ```--stat``` results.

danpoltawski:
This is kind of awkard in where to put it because it sits more on the 'shell' side than the php side. I suppose we could pass a list of changed files to the php, not sure if that a good idea.

`git diff-tree --no-commit-id --name-only -r <commit-sha> | tr '\n' ','`
to do that list

danpoltawski:
I've added some commits [implementing this file deetcion](https://github.com/moodlehq/moodle-local_ci/pull/129/commits/463f659985c425dbf303a876d681cac6993df693)

stronk7:
Seems bad amos script does need amending now, cz the missing lang file problem is getting precedence?

danpoltawski:
Yes, just updated the remote fixture (and rebased this)

stronk7:
Here we go! Let's break the world!

PR #133, opened 2016-08-26
MDLSITE-4704 commit tests: switch to git rev-list

danpoltawski:
It's what we are using in the actual code.

### DIFF
--- a/remote_branch_checker/remote_branch_checker.sh
+++ b/remote_branch_checker/remote_branch_checker.sh
@@ -216,7 +216,7 @@ set -e
 # The merge succeded, now push our precheck branch for inspection:
 if [ ! -z ${pushremote} ]; then
     echo "Info: Pushing precheck branch to remote"
-    pushbranchname=${issue}-${integrateto}-$(git rev-parse --short HEAD)
+    pushbranchname=${issue}-${integrateto}-$(git rev-list -n1 --abbrev-commit HEAD)
     $gitcmd push $pushremote ${integrateto}_precheck:${pushbranchname}
 fi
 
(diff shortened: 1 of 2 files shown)


stronk7:
There is one more occurrence in [the prechecker](https://github.com/stronk7/moodle-local_ci/blob/master/remote_branch_checker/remote_branch_checker.sh#L219). Maybe we should kill that one too? Heh, and add test looking for occurrences in the future?

danpoltawski:
Added a patch for that, not easily testable so here it from a manual test:
```
Info: Pushing precheck branch to remote
Counting objects: 396, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (138/138), done.
Writing objects: 100% (396/396), 87.78 KiB | 0 bytes/s, done.
Total 396 (delta 301), reused 338 (delta 248)
To git@git.in.moodle.com:integration/prechecker.git
 * [new branch]      master_precheck -> MDL-55372-master-master-6bb410c
```

With regard to tests to prevent it, I think that it would be hard to add a test for that because you can name the arguments however you like. But maybe we should be getting rid of all uses of rev-parse because of what you say about rev-list being a porcelain command?

stronk7:
Ah, I was thinking about something static like we did [here](https://github.com/stronk7/moodle-local_ci/blob/master/tests/00-verifications.bats#L15), so any ' --short ' is simply banned in local_ci.

stronk7:
(and I already looked for occurrences of "git rev-parse", the 2 fixed in this PR were the only ones present, unless i'm wrong). So goal achieved.

stronk7:
So,

1) Or we just ban ' --short ' (current PR)
2) Or we ban all ' rev-parse ' uses (or force them to be verified using ````HASH_OR_TAG^{commit}```` ). Surely I prefer complete ban if ' rev-list ' gives us them alredy verified.

stronk7:
(sorry for all the editions, grrr, it seems my brain and my mouse are totally unconnected today)

stronk7:
So are you happy with 1), aka, current patch r(ban ````--short```` and forget).

danpoltawski:
Yes, if we want to do the rest we can do it in a future issue.

stronk7:
merged, thanks!

PR #134, opened 2016-08-30
MDLSITE-4722 take2 - Make old branches to work

stronk7:
As [commented @ MDLSITE-4722](https://tracker.moodle.org/browse/MDLSITE-4722?focusedCommentId=424622&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-424622), I was performing the final tests when I detected that branches < 31_STABLE were failing. The reason is that the ````database_column-info```` internals are different in those old branches.

So this commit just changes the introspection to work differently for 31_STABLE and up ($data protected property) and older branches (individual public properties).

I've already merged this into laptop, after passing local tests, just to be sure it solves the problem. And it seems to be working ok:

- Before (failing): [link](http://ci7.stronk7.com/view/M27/job/25.%20Compare%20databases%20from%2022-23-24-25-26%20(27_STABLE)/1177/console)
- After (passing): [link](http://ci7.stronk7.com/view/M27/job/25.%20Compare%20databases%20from%2022-23-24-25-26%20(27_STABLE)/1178/console)

This includes the fix and a new bats tests to cover the use of old branches.

Ciao :-)

### DIFF
--- a/compare_databases/compare_databases.php
+++ b/compare_databases/compare_databases.php
@@ -164,9 +164,15 @@ function compare_column_specs($tname, $cname, $specs1, $specs2) {
     if (!isset($validelements)) {
         // Use reflection to find all the magic properties of the object.
         $reflection = new ReflectionClass($specs1);
-        $property = $reflection->getProperty('data');
-        $property->setAccessible(true);
-        $validelements = array_keys($property->getValue($specs1));
+        if ($reflection->hasProperty('data')) {
+            // Moodle 3.1 and up, everything is handled magically within the data property).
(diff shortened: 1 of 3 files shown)


stronk7:
I've added a 3rd commit, in order to ensure the origin remote information is updated (fetched) and ensuring any origin branch/local branch/tag or hash always work. Here I faced some problems with my local git clone missing some branches and leading to error, so better we look both locally and in remote. Tests do cover, now, branch, tag and commit usages.

danpoltawski:
Thanks Eloy, merged

PR #135, opened 2016-08-31
MDLSITE-3688 - detect lack of MOODLE_INTERNAL

danpoltawski:
Created the pull request to discuss the idea/approach. I thought this was going to be a nice way to do it, taking a slightly hacky approach to 'abusing' the php -l check, because it would be filtered by the diff excluder by using 'on line 1' and look at the code, we don't have to add much to achieve it.

But now i'm not so sure, because it'd break the post-checks if any file was modified without a MOODLE_INTERNAL.. I suppose we could potentially add a flag for whether it runs the MOODLE_INTERNAL check, but maybe this is just too much of an abuse.

### DIFF
--- a/php_lint/php_lint.sh
+++ b/php_lint/php_lint.sh
@@ -70,6 +70,20 @@ for mfile in ${mfiles} ; do
                 echo "$fullpath - ERROR: BOM character found"
                 errorfound=1
             fi
+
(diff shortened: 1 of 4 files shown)


danpoltawski:
Closing in preference to https://github.com/moodlehq/moodle-local_codechecker/pull/33 which is clearly the better place for it.

PR #136, opened 2016-10-03
MDLSITE-4767 Ignore CSV files in tests

andrewnicols:
CSV files should not be tested for things like illegal whitespace.

### DIFF
--- a/define_excluded/define_excluded.sh
+++ b/define_excluded/define_excluded.sh
@@ -80,6 +80,7 @@ webservice/amf/testclient/AMFTester.mxml
 webservice/amf/testclient/customValidators/JSONValidator.as
 work/
 yui/build/
+*.csv
 *.gif
 *.jpg
 *.png


danpoltawski:
We need to stop this breaking travis

danpoltawski:
We also have an individual csv file in the list already

andrewnicols:
In my defence, Travis hadn't run yet when I opened the PR and I haven't received a failure notification.
Why are we not ignoring all CSVs?

danpoltawski:
@stronk7  I thought this in 2f7f6a0102b130df27d1d558ef5e78b7bd072c69 - I feel a bit like our tests for define_excluded are too heavy-weight now - having to update 6 fixture files every time we need to add one of these is a pain. Maybe we need to change the test to try some sort of exlucuded test in each of the formats?

andrewnicols:
So, do we want to list just that one file, or all csv files?

danpoltawski:
> So, do we want to list just that one file, or all csv files?

I think updating it for all csv files is find (best) - just suggest we remove that manual rule

danpoltawski:
Although, maybe the compare_database failure is showing us this is having unintended cosneququences?

danpoltawski:
I've created #138 to try and fix travis - lets see what the compare database test does

danpoltawski:
Closed by #138 (where I added the tests too)

PR #138, opened 2016-10-03
MDLSITE-4767 Ignore CSV files in tests

danpoltawski:
I couldn't easily collaborate on #136 because its Andrew branch, so creating this one for travis to run..


danpoltawski:
This didn't fix the problem we were trying to fix

PR #139, opened 2016-10-03
NOBUG: Fix regexp variant of excluded to use proper regexp

stronk7:
Apparently it was using some sort of * shell expansion, hence not
working always. Now * is translated to proper .* regexp.

As far as now *.csv are being excluded, we don't need individual
lib/tests/fixtures/tabfile.csv anymore.

Also get rid of mega-old "search" exclude that was causing lots of
new search stuff not to be checked.

And, finally, also "node_modules" that normally CIs don't face, but
can be a source of problems/confusion when working locally.

Of course, with extra commit to fix bats tests (we really should test `01-define_excluded.bats`differently.


stronk7:
WAIT till laptop CIs (the fix has been applied there) run the illegal-whitespace job for master. This is a test till then:

- http://ci.stronk7.com/view/master/job/02.%20Illegal%20whitespace%20(master)/
- http://ci7.stronk7.com:81/view/master/job/02.%20Illegal%20whitespace%20(master)/

If they are back to pass, then the fix is ok.

stronk7:
Ok, with the 2 laptop jobs already passing, together with travis, I'm self-merging this. That should cause public CI server to start passing:

https://integration.moodle.org/view/master/job/02.%20Illegal%20whitespace%20(master)/

stronk7:
Merged, thanks!

stronk7:
ty!

PR #143, opened 2016-10-10
NOBUG: Avoid the job to modify/fetch new stuff

stronk7:
Basically current git status roots the execution.

Looking to those fetches added (recently!) to the database compare
job... i think maybe I added them because of ci tests not ensuring
everything is available in the moodle.git clone (LOCAL_CI_TESTS_GITDIR).
I think I needed it because I was pointing to ongoing integration.git
clone (some tests were dependent of stuff being integrated that week).

I thinkI'm going to, simply, kill the fetches because it's not the job
responsibility to ensure that. travis already does full clone and the
git jobs in jenkins do fetch everything (or can be configured for that f
missing).

### DIFF
--- a/compare_databases/compare_databases.sh
+++ b/compare_databases/compare_databases.sh
@@ -64,9 +64,6 @@ if [ $exitstatus -ne 0 ]; then
     exit $exitstatus
 fi
 
-# Ensure we have all branches and tags at hand or some origin objects may not be available
-cd $gitdir && $gitcmd fetch --all && $gitcmd fetch --tags
-
 # Do the moodle install of $installdb
 echo "Info: Installing Moodle $gitbranchinstalled into $installdb" | tee -a "${logfile}"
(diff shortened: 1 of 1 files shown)


stronk7:
waiting travis to pass and laptops to run it (already updated)...

stronk7:
All seems ok in laptops too:

- http://ci.stronk7.com/job/25.%20Compare%20databases%20from%2027-28-29-30-31%20(master)/3852/

and so on.

danpoltawski:
Merged, thanks!

stronk7:
Just happened again, and I did update laptops before this was merged to master:

http://ci.stronk7.com/job/25.%20Compare%20databases%20from%2027-28-29-30-31%20(master)/3862/

(changes reported - MDL-36611 - do not match the situation being compared by the job, that is using newer commits belonging to MDL-52051 and still should not be there in the CI clone)

So maybe there is some other place where we are fetching changes incorrectly.

PR #144, opened 2016-10-14
MDLSITE-4770 mustache_lint: load partials

danpoltawski:
First commit is the (broken) test which we are going to fix ;-)


stronk7:
Thanks, merged!

PR #145, opened 2016-10-19
MDLSITE-4770 - Make html validator errors be reported to prechecker

danpoltawski:


stronk7:
fair enough, non controversial. So merged, thanks!

PR #146, opened 2016-11-02
Properly detect 'full html body' templates

danpoltawski:
* There was a logic bug (the arguments to strpos the wrong way round!)
* We didn't se tthe content when full body pages were found oops!


danpoltawski:
Will be seen on the cibot results for https://tracker.moodle.org/browse/MDL-56582

stronk7:
Merged, thanks! And already forced update in CIs.

PR #147, opened 2016-11-02
MDLSITE-4787 mustache_lint: Handle theme templates better

danpoltawski:
We have theme templates, which load partials which override the core partials and thus, we are validating incorrect results. At first I thought we didn't need to care about themes and that it was too complex, but we can work out which theme is wanted from the path of the template.


danpoltawski:
You can test this with a real example on integration master:

```
php mustache_lint.php --validator=https://html5.validator.nu/ \
--basename=/path/to/integration_master \
--filename=/path/to/integration_master/theme/boost/templates/core_form/element-static.mustache
```

stronk7:
Merged, thanks. (also to laptop and Ci servers updated)

PR #148, opened 2016-11-02
MDLSITE-4788 mustache_lint: add missing helpers

danpoltawski:
I know its horrible that i'm re-implmeneting all these helpers.

I am not convinced by all of whats going on in the quote helper, and I commented on MDL-56054 about it. But I could just be misunderstanding and in any case, this linter should do what core does anyway

### DIFF
--- a/mustache_lint/mustache_lint.php
+++ b/mustache_lint/mustache_lint.php
@@ -81,6 +81,13 @@ $mustache = new Mustache_Engine([
     'helpers' => [ // Emulate some helpers for html validation purposes.
         'str' => function($text) { return "[[$text]]"; },
         'pix' => function($text) { return "<img src='pix-placeholder.png' alt='$text'>"; },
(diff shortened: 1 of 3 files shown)


danpoltawski:
This was discovered in MDL-55841, so like the previous pull request you can test the problem/fix with:
```
 php mustache_lint.php --validator=http://localhost:8888 \
--basename=/path/to/integration_master \
--filename=/path/to/integration_master/enrol/lti/templates/copy_grid.mustache
```

danpoltawski:
For the record. I am not sure its a wise idea having the uniqid helper in mustache as it is now because it'll return the same value every time. I think that the naming gives a false impression of that. It's unque per template run, but not when including partials and so on.

stronk7:
Merged, pulled to laptop branch and upgraded Ci sites... thanks!

PR #149, opened 2016-11-03
MDLSITE-4790 mustache_lint: Lint JS in templates with eslint

danpoltawski:
It's not possible to do this easily with grunt because it needs the post-rendered output, but we can fit this in here.

### DIFF
--- a/mustache_lint/js_helper.php
+++ b/mustache_lint/js_helper.php
@@ -1,0 +1,86 @@
+<?php
+
+/**
+ * Helper to process JS and lint it.
+ */
+class js_helper {
+    protected $js;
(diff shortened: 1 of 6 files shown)


danpoltawski:
Added tests, ready for merging

danpoltawski:
After last night's discussion, I have updated the branch with 46d7936. See the commit message for the explanation.

Note, we could go one step further and disable the html/eslint checks completely when example context is missing. I prefer not to do that because doing it then way gives more of an incentive to fix the example context and also can spot problems which will go missed otherwise. These are all warnings.

stronk7:
yay-thanks

PR #152, opened 2016-11-10
NOBUG: filter skip-ci-version-check a better way

danpoltawski:
The previous method of limiting the results returned from the tracker
meant that we would then falsely report about these commits when looking
for unknowned commits.


stronk7:
I've proceeded to merge it as far as the travis failures seem to be some recent change in ordering or so.

PR #155, opened 2016-11-16
MDLSITE-4806 remove the "ci" label when issues arrive to integration

stronk7:
When an issue, having the ci label arrives to "waiting for integration"
it's interesting to get the label cleaned, so the prechecker can verify
that everything is correct against current upstream.

This is specially noticeable in continuous-integration periods where
the issues are never moved to "current integration" (monday automated
job) so last prechecker results are really old (happened normally before
peer-review).

This job picks those issues and cleans the label, forcing an extra
prechecker to happen. Has built in machinery to prevent the same issue
to be cleaned too many times (in 1 hour) and other details. Should be
safe to be executed at any time interval.

### DIFF
--- a/tracker_automations/remove_ci_label_from_waiting_integration/remove_ci_label_from_waiting_integration.sh
+++ b/tracker_automations/remove_ci_label_from_waiting_integration/remove_ci_label_from_waiting_integration.sh
@@ -1,0 +1,92 @@
+#!/usr/bin/env bash
+# Look for all the issues awaiting integration that were not awaiting integration the last time
+# the job was executed and having the "ci label in order to remove it and get prechecker performed.
+#jiraclicmd: fill execution path of the jira cli
+#jiraserver: jira server url we are going to connect to
+#jirauser: user that will perform the execution
+#jirapass: password of the user
+
+# Let's go strict (exit on error)
+set -e
+
+# Verify everything is set
+required="WORKSPACE jiraclicmd jiraserver jirauser jirapass"
+for var in $required; do
+    if [ -z "${!var}" ]; then
+        echo "Error: ${var} environment variable is not defined. See the script comments."
+        exit 1
+    fi
+done
+
+# file where results will be sent
+resultfile=$WORKSPACE/remove_ci_label_from_waiting_integration.csv
+echo -n > "${resultfile}"
+
+# file where updated entries will be logged
+logfile=$WORKSPACE/remove_ci_label_from_waiting_integration.log
+lastfile=$WORKSPACE/remove_ci_label_from_waiting_integration_latest.log
+
+# Calculate some variables
+mydir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+basereq="${jiraclicmd} --server ${jiraserver} --user ${jirauser} --password ${jirapass}"
+BUILD_TIMESTAMP="$(date +'%Y-%m-%d_%H-%M-%S')"
+now=$(date +%s)
+
+# Note this could be done by one unique "runFromIssueList" action, but we are splitting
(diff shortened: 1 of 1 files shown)


stronk7:
I'm going to install this in the laptop CI and play a bit with it and your MDL-56922 issue, to see how it flows over different situations. Stay tuned for results confirmation.

tracker_automations/remove_ci_label_from_waiting_integration/remove_ci_label_from_waiting_integration.sh
```diff
@@ -0,0 +1,92 @@
...
+for var in $required; do
+    if [ -z "${!var}" ]; then
+        echo "Error: ${var} environment variable is not defined. See the script comments."
+        exit 1
+    fi
+done
+
+# file where results will be sent
+resultfile=$WORKSPACE/remove_ci_label_from_waiting_integration.csv
+echo -n > "${resultfile}"
```
danpoltawski: Out of interest, is there a reason for not using 'touch' here?

tracker_automations/remove_ci_label_from_waiting_integration/remove_ci_label_from_waiting_integration.sh
```diff
@@ -0,0 +1,92 @@
...
+        fi
+    fi
+    if [[ -n $canremovelabel ]]; then
+        ${basereq} --action removeLabels \
+           --issue ${issue} \
+           --labels "ci"
+        echo ${issue} $(date +%s) >> "${lastfile}"
+        echo "    - 'ci' label removed"
+    fi
+    echo "$BUILD_NUMBER $BUILD_TIMESTAMP ${issue}" >> "${logfile}"
```
danpoltawski: I don't understand what the point of this logfile is (compared to the console logs), it doesn't seem to add anything? (It was adding confusion to me because of the two different timestamp formats we're using in the two different files).

tracker_automations/remove_ci_label_from_waiting_integration/remove_ci_label_from_waiting_integration.sh
```diff
@@ -0,0 +1,92 @@
...
+done
+
+# Auto-clean the lastfile when possible. If the last entry in the file
+# is older than 1 hour, it can be safely deleted because not entry in
+# it will match the conditions above anymore.
```
danpoltawski: Ah, clever. So there are periods where no issues are sent for integration and then we can clean the file so its not self-pepetituating

tracker_automations/remove_ci_label_from_waiting_integration/remove_ci_label_from_waiting_integration.sh
stronk7: there is not good reason really. I think that, along the years, I only have used them a couple of times, if for any reason I had to look about when something happened in a given issue.

I'm happy getting rid of all them, but in another issue. Maybe together with previous one.

tracker_automations/remove_ci_label_from_waiting_integration/remove_ci_label_from_waiting_integration.sh
stronk7: yeah, i was over-complicating the cleanup algorithm till, suddenly, realized that it's 100% safe to delete the file once all records (last one s checked) are > 1h old.

tracker_automations/remove_ci_label_from_waiting_integration/remove_ci_label_from_waiting_integration.sh
danpoltawski: +1

danpoltawski:
Merged, thanks!

PR #156, opened 2016-11-18
NOBUG: set timeouts for curl request

danpoltawski:
Just in case it was the cause of
https://integration.moodle.org/job/Precheck%20remote%20branch/29665/
being stuck for 9 hours


stronk7:
Merged before tests ended (precheck related one already ran, I think). :-P

PR #157, opened 2016-11-21
Attempt to parallelise the travis build

danpoltawski:
Creating this pull request but realise that the test naming is not ideal, so want to discuss that here. Maybe should just get rid of the 01, 99 and so on and just have the 1- and 2- prefix (and future prefixes for future splits up). 

Any thoughts?


stronk7:
Just guessing if a possible alternative could be to have the suites spread into subfolder (suite1, suite2...) and then, simply, the tests within them, without any numbering at all (after all they are not important, or the suites pass or they fail as a whole, so we always have to wait, and for quick testing, always can create a "suite3" temp subfolder with the new tests or whatever).

That way, it would be really easy to "balance" suites, just moving tests within folders.

Just my POV, ciao :-)

stronk7:
(basically my point is that any prefix (suite number or test order) are distracting)

danpoltawski:
Yep, I like the idea of folders. Only concern is that there are some things where we depend on paths.

danpoltawski:
I updated the patch to just add the one prefixed number. I know you mentioned the suffix, but when I was about to do it, I didn't know what advantage it had (and also thought it could be more easily confused with two sets of the same tests)

stronk7:
(the only advantage of the suffix is that you don't have to look alphabetically twice to find a given test, that's the only reason i proposed it. Anyway with the old prefixes gone, perfect. Just 2 vs N old ones)

stronk7:
Merged, thanks!

PR #158, opened 2016-11-21
MDLSITE-4625 Report grunt errors in the prechecker

danpoltawski:
Previously the prechecker was only reporting on undetected changes in files from grunt. This makes it report on grunt failures.

### DIFF
--- a/remote_branch_checker/checkstyle_converter.php
+++ b/remote_branch_checker/checkstyle_converter.php
@@ -110,6 +110,13 @@ function process_gruntdiff($line) {
         $output.= '<error line="0" column="0" severity="error" ';
         $output.= 'message="Uncommitted change detected."/>' . PHP_EOL;
         $output.= '</file>';
+    } else if (preg_match('/^ERROR: (.*)$/', $line, $matches)) {
+        $error = $matches[1];
+
+        $output.= '<file name="">' . PHP_EOL;
+        $output.= '<error line="0" column="0" severity="error" ';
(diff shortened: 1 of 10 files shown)


stronk7:
Only point I can imagine is that "GRUNT-CHANGE" output comes also with an "ERROR: Some modules are not properly ..." line that now will be reported, together with the changes and previously was not. I've seen you've adapted that in the tests. If you think it's ok enough, I'm happy.

danpoltawski:
> Only point I can imagine is that "GRUNT-CHANGE" output comes also with an "ERROR: Some modules are not properly ..." line that now will be reported, together with the changes and previously was not. 

I think thats good, no?

stronk7:
I'm not sure if it's good (not saying it's bad either, heh). Just it's like a redundant error when after it the detailed (GRUNT-CHANGE) cases (with file...) come.

stronk7:
There are 2 alternative:

- or we leave it as is. And the generic, colon terminated error is shown before the GRUNT-CHANGE occurrences.
- or we change it to "PROBLEM" and then it's not reported (because the GRUNT-CHANGE occurrences are enough).

I'm happy with both.

danpoltawski:
For me if grunt is failing, that is **more important** to display the real error from grnt than the GRUNT-CHANGE (which is just a report of the final outcome). So I think its fine to have both.

stronk7:
Oki, merging then...

PR #159, opened 2016-11-23
NOBUG: Add 2 DOS-EOL fixtures to get the test back passing

stronk7:
I've confirmed locally that the count is back to ` current count: 959` with the templates out.


stronk7:
Amended to get related tests passing:

````
$ ../../../bats/bin/bats tests/01-define_excluded.bats
 ✓ define_excluded: generates results correctly for all formats
 ✓ define_excluded: is immune to trailing slashes in gitdir

2 tests, 0 failures
````

https://tracker.moodle.org/browse/MDLSITE-4819 should be next...

danpoltawski:
Timeout issue, looks like #157 is more urgent, but lets get this landed first

PR #161, opened 2016-11-24
NOBUG: Avoid slashed LOCAL_CI_TESTS_GITDIR, breaks various tests

stronk7:
It already has happened twice (see #111) and it's a pain to
realize / detect. Initially I was to going to incorporate this
as a test to setup.bats, but now that the numbering is gone
that test is not executed the first anymore.

So instead, I just clean any trailing slash and forget.


danpoltawski:
Merged, thanks!

PR #162, opened 2016-11-24
NOBUG: remove the SMURFRESULT assertion from prechecker tests

danpoltawski:
The problem is that when we assert the smurf result in the console
output we are just duplicating the assertion which is already happening
by comparing the .xml file. But because we fail before looking at the xml
file we don't get to see how the output differs.. which means in the
case of random travis errors it makes it hard for us to diagnose what
the problem is.

Probably best just to remove that duplciated assertion and hopefully
we'll be able to more easily work out whats going on in case of
failures.


stronk7:
Fair enough, merged (btw this time dear Travis passed)

PR #163, opened 2016-11-24
MDLSITE-4819 tests: Better define_excluded tests

stronk7:
Now it's easier to handle the fixtures as far as,
instead of comparing the whole fixture, each line is
checked to exist in results independently.

I've left the default format (format-excluded-expected.txt)
unmodified so, at least, that one contains all current results
but checked line by line. Other formats now just have a smaller
set being checked.


danpoltawski:
Merged, thanks

danpoltawski:
Grr, just realised this pull request was against laptop! Sending to master

danpoltawski:
grr, i can't, non-ff because of #162 merge not being spread

danpoltawski:
laptop branch history rewritten

stronk7:
oh, man sorry. I'll start using feature branches instead of laptop (using it only for testing purposes). bad habit, specially when I always forget to set the PR properly.

PR #164, opened 2016-11-25
MDLSITE-4808 prechecker: Inform to other jobs about checked issues

stronk7:
With a new optional parameter, informtofiles, where a list of
files can be specified. The job will write one line to each of those
files with format:

MDL-xxxxxxx  unixtime bulk_precheck_issues

Just that, it's the responsibility of other jobs to read and
process that information properly.

### DIFF
--- a/tracker_automations/bulk_precheck_issues/bulk_precheck_issues.sh
+++ b/tracker_automations/bulk_precheck_issues/bulk_precheck_issues.sh
@@ -8,6 +8,7 @@
 #cf_branches: pairs of moodle branch and id for "Pull XXXX Branch" custom field (master:customfield_10111,....)
 #cf_testinginstructions: id for testing instructions custom field (customfield_10117)
 #criteria: "awaiting peer review", "awaiting integration", "developer request"
+#informtofiles: comma separated list of files where each MDL processed will be informed (format MDL-xxxx unixseconds)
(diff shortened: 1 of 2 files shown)


stronk7:
WIP, WIP, WIP, do not merge yet, WIP, WIP, WIP

stronk7:
Seems to be working already. See [these comments](https://tracker.moodle.org/browse/MDLSITE-4808?focusedCommentId=443149&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-443149).

It's already in the laptop branch and running from there. Once this gets accepted to master, I'll configure the jobs in public and switch to it.

So, ready at all effects for your review.

Ciao :-)

danpoltawski:
Merged, I am not completely sold on this approach to be honest as I feel like its not ideal having the two different processes writing to the same file, there could be race conditions. But, I think overall that is solved by the brain for cibot [MDLSITE-4807](https://tracker.moodle.org/browse/MDLSITE-4807)

stronk7:
Yeah, to be honest I thought about that and was near, near adding locking. But as far as they are micro-operations (happening with >= 10 minutes frequency) finally decided to keep it that way.

Sure once we have a brain (file/db backed) that will allow better intercommunication techniques.

PR #167, opened 2017-01-31
HQ-1428 groovy script for adding notificaitons to all jobs

danpoltawski:
Intended to be run in the script console with some rules about how many log lines to send for each job. We can use this to tweak the different job notifications as we find out what output is useful or not.

(Note that I hope we won't need to use this often because we'll convert to use pipelines which makes things like this way better, its really really worth us doing it I think, will make maintenance of our jobs much easier).


stronk7:
Thanks, I've merged this locally after adding a little modification to let bats tests to know about .groovy extension:

https://github.com/moodlehq/moodle-local_ci/commit/8a22f3e6b4b17b6f8a6e01bb684ecf87badde0c6#diff-09f71f4851bf6f8c63dac1be31409f05L26

Was tested on laptop CI servers and worked ok, so feel free to remove the candidate branch.

Ciao :-)

PR #168, opened 2017-02-01
Notification adjustments

danpoltawski:
* Don't notify on the prechecker job (this is expected to fail
  and reports to the tracker)
* Always inform about failures on the security rebase and check
  mark as integrated jobs


stronk7:
Merged, thanks!

PR #171, opened 2017-02-17
NOBUG: Always delete any _precheck branch from previous executions

stronk7:
We cannot delete the _precheck branches while they run because that
will cause all the /work to be lost before being saved as artifacts.

So, on every run, before performing any other git operation, we check
if we are in a _precheck branch and, if so, proceed to checkout its
base counterpart, checkout it and delete the _precheck branch and
any old stuff it may have.

Note this was introduced because we found some plugins having own
.gitattributes files, playing with auto-crlf settings and causing
next executions of the checker to fail forever. Now, we ensure
nothing is left in the prechecker repo from previous executions.

### DIFF
--- a/remote_branch_checker/remote_branch_checker.sh
+++ b/remote_branch_checker/remote_branch_checker.sh
@@ -70,6 +70,17 @@ if ! $(git remote -v | grep -q '^integration[[:space:]]]*git:.*integration.git')
     ${gitcmd} remote add integration git://git.moodle.org/integration.git
 fi
 
+# We are into a _precheck branch from previous run. We don't like any leftover
+# so let's delete it drastically. Will be recreated later if needed.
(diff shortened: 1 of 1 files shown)


stronk7:
(let's see if the extra Info does break any test)

stronk7:
Tested with the original case that caused the problems.....

I'm running the problematic plugin again, so the repo is left in unstable status (for next run):

https://integration.moodle.org/view/prechecker/job/Precheck%20remote%20branch/32969/

So, the problem is back for all next runs:

https://integration.moodle.org/view/prechecker/job/Precheck%20remote%20branch/32970/

And now, with this PR applied (it has passed travis tests and I've merged it)... jobs are back to work, because the problematic _precheck branch from previous execution is being deleted:

https://integration.moodle.org/view/prechecker/job/Precheck%20remote%20branch/32971/

PR #172, opened 2017-02-20
MDLSITE-4892 deal with spaces in passed varibles

danpoltawski:
Because this comes from tracker fields, we can trim the spaces. Probably
would be more appropiate in the remote branch checker, but its better
covered with tests here and doesn't do any harm.

I decided not to add a testcase for the 'ok with space' case to avoid
wasting the time of a full precheck run.

### DIFF
--- a/remote_branch_checker/remote_branch_checker.sh
+++ b/remote_branch_checker/remote_branch_checker.sh
@@ -146,6 +146,11 @@ basecommit=$(${gitcmd} rev-parse --verify ${baseref})
 # (NOTE: checkout -B means create if branch doesn't exist or reset if it does.)
 ${gitcmd} checkout -q -B ${integrateto}_precheck $baseref
 
+# Do some cleanup onto the passed details
+
+# Trim whitespace in branch/remote
+remote=${remote//[[:blank:]]/}
+branch=${branch//[[:blank:]]/}
 
 # Fetch the remote branch.
 set +e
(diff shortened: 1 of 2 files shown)


remote_branch_checker/remote_branch_checker.sh
```diff
@@ -146,6 +146,11 @@ basecommit=$(${gitcmd} rev-parse --verify ${baseref})
 # (NOTE: checkout -B means create if branch doesn't exist or reset if it does.)
 ${gitcmd} checkout -q -B ${integrateto}_precheck $baseref
 
+# Do some cleanup onto the passed details
+
+# Trim whitespace in branch/remote
```
danpoltawski: Yeah. I think, my initial idea was to trim, but it was easier just to get rid of all spaces.

* As it needs to be a url the remote can't have spaces
* neither can branch: https://www.kernel.org/pub/software/scm/git/docs/git-check-ref-format.html

tests/2-remote_branch_checker.bats
```diff
@@ -101,3 +101,16 @@ assert_prechecker () {
 @test "remote_branch_checker/remote_branch_checker.sh: grunt build failed" {
     assert_prechecker fixture-grunt-build-failed MDL-12345 cd4a6b8b0bca159d3abb1468794ed5a074c5b701
 }
+
+@test "remote_branch_checker/remote_branch_checker.sh: remote which doesnt exist" {
+    export branch="a-branch-which-will-never-exist"
```
danpoltawski: I did this because I am not testing the new code, I am testing the existing 'remote doesn't exist' code. Because thats what sort of surprised me about the way this stuff was not failing with a better error, when we had code to detect bad remotes. 

 I thought the test for the trimming the spaces would be a little bit misleading to adding it in.

stronk7:
All right, then. Merged, thanks!

PR #173, opened 2017-02-20
MDLSITE-3758 sort out bad response to incorrect branch name

danpoltawski:
MDLSITE-3758 sort out bad response to incorrect branch name


danpoltawski:
rebased.

danpoltawski:
It's been so long since I looked at this, I forget exactly what was happening.

stronk7:
> It's been so long since I looked at this, I forget exactly what was happening

The code looks ok, trying to extract the branch name from the url, because it seems to be a recurring mistake to put URLs instead of branch names is those fields.

But your comment above is what has confused me. What did to exactly forgot?

danpoltawski:
Referring to the first commit where the 'is valid remote branch' check was failing.

stronk7:
Aha, got it now. Thanks, merged !

PR #174, opened 2017-04-03
Add ability to use the VNU jar file

mrmark:
Instead of calling the HTML validator
for Mustache lint as a service, call
it as a JAR file.

### DIFF
--- a/mustache_lint/mustache_lint.php
+++ b/mustache_lint/mustache_lint.php
@@ -43,6 +43,7 @@ Options:
 -h, --help            Print out this help
 -f, --filename        Path to file
 -v, --validator       Validator URL (e.g. https://html5.validator.nu/ or http://localhost:8080/)
+                      or path to validator JAR file (e.g. /path/to/vnu.jar)
 -b, --basename        Full path to the moodle base dir
 ";
     echo $help;
@@ -252,6 +253,26 @@ function print_eslint_problems($problems, $ignoreparseerrors) {
 function validate_html($content) {
     global $VALIDATOR;
 
+    if (strpos($VALIDATOR, 'http') !== 0) {
+        $file = tempnam(sys_get_temp_dir(), 'mustache_lint');
+
+        if ($file === false) {
+            return false;
+        }
+        file_put_contents($file, $content);
+
+        // The "2>&1" part redirects stderr to stdout because the JSON is sent to stderr and PHP does not capture that.
+        $output = shell_exec(sprintf('java -jar %s --format json --exit-zero-always %s 2>&1', $VALIDATOR, $file));
+
+        unlink($file);
+
+        if ($output === null) {
+            return false;
+        }
+
+        return json_decode($output);
+    }
+    
     $ch = curl_init("$VALIDATOR?out=json");
     curl_setopt($ch, CURLOPT_USERAGENT, 'moodle/local_ci validator 0.1');
     curl_setopt($ch, CURLOPT_HTTPHEADER, ['Content-type: text/html; charset=utf-8']);


mrmark:
Hey @danpoltawski - what do you think about this for https://tracker.moodle.org/browse/MDLSITE-5003 ?  This solution would also work better for `moodle-plugin-ci` as I would have to use some `killall` trickery to kill the JAVA server because Travis hangs if you leave any background processes running.

Also, is there a way to download the JAR file in this script or should I handle that myself?  Not a big deal either way, but would ensure we use the same version.

danpoltawski:
Thanks Mark! I don't think I noticed that it was possible to call vnu.jar like this, otherwise I think i would've supported it like that in the first place, will be much better for that MDLSITE. So this looks perfect.

I don't think we have a good way to install the jar file, I can't decide what would be the best way to do it, shame there isn't a composer wrapper for it..

mrmark:
Have you looked at `npm install vnu-jar`?  There is also a `pip` way but I'm not sure if I can install that in Travis when it is a PHP built.

mrmark:
FYI currently manually testing this - the output from the JAR is going to stderr and not getting captured by PHP exec type commands.

mrmark:
Ouch, just noticed that the JAR requires JAVA8 which isn't installed by default.  Adding a lot of complexity to the build process here in Travis.  Do you know how risky it is for me to use `https://html5.validator.nu/` in Travis CI?  EG: rate limits?

danpoltawski:
I think its not going to be sustainable (our jenkins install seems to be blocked from it now)

mrmark:
So, after a lot of digging, this appears to get the JDK into container based builds:
```yaml
addons:
  apt:
    packages:
      - oracle-java8-installer
      - oracle-java8-set-default
```
This doesn't seem as bad and worth doing to get the linting.

FYI, manual testing worked and I updated the PR with the fixes.  Let me know if there is anything else I can do to help get this integrated.

mrmark:
I tried out `npm install -g vnu-jar` and the JAR file is available in `/usr/local/lib/node_modules/vnu-jar/build/dist/vnu.jar` - so that seems like a good option, I might stick with that to keep things up-to-date.

Just a link/brain dump:
* The NPM package is provided by the project itself:
  * https://github.com/validator/validator/blob/master/npm.md
  * https://github.com/validator/validator/blob/master/package.json
  * https://github.com/validator/validator/blob/master/vnu-jar.js
* There is a Grunt tool for this, but I'm guessing migrating to it would be hard: https://github.com/jzaefferer/grunt-html

danpoltawski:
I was a bit hesitant about the npm route, because its yet another package manager dependency on this repo, but I am coming round to it

a) It's the path of least resistance 
b) It's not critical part of setup to do the npm install, you can just run it when you start using this validation feature and want to use a local validator
c) We can easily make use of it in our jenkins instlal this way

Plus we can then add this to the bats tests. So I think i'll prepare some commits to add the packge.json and tests

mrmark:
I have it working in v2 of MPCI - its quite nice and it can be cached since it is an NPM package, which is great since the JAR is quite large.

I did find a trick to locating the JAR file, if you run `npm -g prefix` then that outputs the install path.  Then just concatenate that path with `/lib/node_modules/vnu-jar/build/dist/vnu.jar`

If you do end up adding a `package.json` and run `npm install`, if at all possible, please exclude it from the `mustache_lint/mustache_lint.php` file, as MPCI would have already installed it.  I'm open to discussion on this of course though.

danpoltawski:
I have put the extra bits in #176, but going to continue the discussion here..

@stronk7 - interested in your thoughts on this:

- Running the validator locally using java seems a great approach for us. 
- Using NPM to lock versions and run tests seems the right solution for us and Mark and other constumers of local_ci..
- However, this java depdency is nothing javascripty..
- Installing the .jar using NPM has some downsides for us, because...

We already have the slightly crazy hard to understand situation with npm in local_ci (see prepare_npm_stuff/prepare_npm_stuff.sh). Mostly thats about local moodle versions on test, but we do still have those legacy shifter/recess commands too..

In the end, I think this is the right way to go, but I am aware we will now have 3 differnet npm locations:

- The 'moodle branch under test' place
- The global legacy shiter/recess place
- This new, npm dependencies for local_ci itself place

I do think this is a good way for this use, but it sure its complicated.

mrmark:
> However, this java depdency is nothing javascripty..

They do have a Grunt plugin, so you could run the validation via Grunt if you wanted.  I'm guessing no though since this is already separated from Moodle.

In regards to the NPM locations, you could keep the `vnu-jar` installed globally.  I'm guessing you were always working with the latest and greatest so far.  I suppose locking the version could be handy at some point, but with some coordination, we could lock it via the install command by installing a specific version.

danpoltawski:
I prefer getting away from the manually installed global dependencies we have, but yeah thats an optio too.

stronk7:
+1 for this / #176 I think it's ok to have it via local composer/npm definitions.

Just... maybe... i'd make the validator checker... a bit more complete, so it not only looks for the existence of the jar file but actually verifies that can be java-executed (maybe asking for --version or whatever). If that does not work, then online one.

Cannot find any drawback, it's clear the new npm stuff are requisites for local_ci and, as such, should be in installation instructions somehow (composer ones too, btw).

But I like it!

danpoltawski:
> Just... maybe... i'd make the validator checker... a bit more complete, 

Ok i've done that. It also needed to detect too-low java version. Here is the commit:
https://github.com/moodlehq/moodle-local_ci/pull/176/commits/42814ddc600c1ee42976c50c65e418d60c613cd2

And you can see what happens here:
https://travis-ci.org/moodlehq/moodle-local_ci/builds/232837145#L296

danpoltawski:
(It was a bit of a pain doing that version check)

stronk7:
I imagine this can now be safely closed with #176 already landed.

PR #177, opened 2017-04-24
Add ANY_VERSION to fix PHP Notice

mrmark:


danpoltawski:
Related to comment I made in https://tracker.moodle.org/browse/MDLSITE-5007 - we really should be seeing these notices..

danpoltawski:
Merged, thanks Mark

PR #179, opened 2017-06-05
MDLSITE-5134 ci: moving to nvm nodejs manager

stronk7:
This does:
- make both npmcmd and nodecmd globals optional,
  defaulting to "npm" and "node" respectively.
  (note there is MDLSITE-4728 about the final
  removal but surely we'll need to branch before
  that happens and delete other old bits like
  shifter/recess)
- better printing the currently used npm/node versions.
- in case grunt-cli needs to be installed explicitly (I've
  tried here with node 4, 6, 7 - npm 2, 3, 5 - and it was
  installed always), it's done with --no-save, so package
  and shrinkwrap files are not modified.
- in prechecker, just moved the mandatory checks before
  applying defaults, to be the same than all other scripts.
- move the 2 "cmds" out from tests as far as they are optional.

### DIFF
--- a/grunt_process/grunt_process.sh
+++ b/grunt_process/grunt_process.sh
@@ -3,7 +3,7 @@
 # $gitdir: Directory containing git repo
 # $gitbranch: Branch we are going to install the DB
 # $extrapath: Extra paths to be available (global)
-# $npmcmd: Path to the npm executable (global)
+# $npmcmd: Optional, path to the npm executable (global)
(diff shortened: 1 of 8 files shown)


stronk7:
DoNoTmErGe (read the issue)!

stronk7:
Note that once this is merged, that [steps defined in the tracker](https://tracker.moodle.org/browse/MDLSITE-5134) may need to be performed, so a friendly ping will be welcome. TIA!

danpoltawski:
Merged, thanks!

PR #180, opened 2017-07-20
MDLSITE-4710 thirdparty_check: Strict filepath match

dmonllao:
It should be able to match modifications in third party libs that are just 1 file and modifications in third party libs directories.


danpoltawski:
Thanks David, merged it - I added a commit testing this edgecase: https://github.com/moodlehq/moodle-local_ci/commit/edddbb45666b87be9e37723ebc877f5f44e8d45c

PR #181, opened 2017-09-26
Detect multiple calls to save point with same version

mrmark:
If a save point is called with the same version, it causes a downgrade error.  This new check looks for duplicate version save point calls throughout the whole upgrade script.

Example of what this catches:
```php
<?php
function xmldb_forum_upgrade($oldversion) {
    if ($oldversion < 2014051201) {
        upgrade_mod_savepoint(true, 2014051201, 'forum');
    }
    if ($oldversion < 2014051201) {
        upgrade_mod_savepoint(true, 2014051201, 'forum');
    }
    return true;
}
```
This causes a downgrade error because you cannot call the save point method with the same or lower version.

This error can be "doubled" up due to the second check that counts the number of save points in each if block.  EG:
```php
<?php
function xmldb_forum_upgrade($oldversion) {
    if ($oldversion < 2014051201) {
        upgrade_mod_savepoint(true, 2014051201, 'forum');
        upgrade_mod_savepoint(true, 2014051201, 'forum');
    }
    return true;
}
```
Would print multiple errors about basically the same thing.  Hopefully that is OK, since downgrade errors are pretty nasty.

### DIFF
--- a/check_upgrade_savepoints/check_upgrade_savepoints.php
+++ b/check_upgrade_savepoints/check_upgrade_savepoints.php
@@ -161,6 +161,15 @@ foreach ($files as $file) {
         }
     }
 
+/// Let's ensure there are no duplicate calls to a save point with the same version.
+    if ($count_sp > 0) {
+        foreach (array_count_values($matches2[2]) as $version => $count) {
+            if ($count > 1) {
+                echo "    + ERROR: Detected multiple 'savepoint' calls for version $version".LINEFEED;
+            }
+        }
+    }
+
 /// Let's split them
     if (!preg_match_all('@(' . $if_regexp . '(\{(?>(?>[^{}]+)|(?3))*\}))@is', $contents, $matches)) {
         echo "    + NOTE: cannot find 'if' blocks within the upgrade function" . LINEFEED;
@@ -196,7 +205,7 @@ foreach ($files as $file) {
             echo "    + ERROR: version $version is missing corresponding savepoint call" . LINEFEED;
             $has_version_mismatch = true;
         } else if ($count_spv > 1) {
-            echo "    + WARN: version $version has more than one savepoint call" . LINEFEED;
+            echo "    + ERROR: version $version has more than one savepoint call" . LINEFEED;
             $has_version_mismatch = true;
         } else {
             if ($version !== $matches[2][0]) {
--- a/tests/1-check_upgrade_savepoints.bats
+++ b/tests/1-check_upgrade_savepoints.bats
@@ -27,9 +27,10 @@ setup () {
 
     ci_run check_upgrade_savepoints/check_upgrade_savepoints.sh
     assert_failure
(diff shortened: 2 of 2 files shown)


stronk7:
Hi @mrmark,

1st, I really thought we were looking already for repeated savepoints, but it seems we only were looking for wrongly ordered in non-strict mode (equal allowed). In any case, I agree your new check is nice and detects those repetitions pretty well.

Then, there are the tests, that are failing with your patch, and passing before. I bet it's because it's taking the pipe added there (`ERROR|WARN`) as a pipe, leading to unfinished work. I've tried here with `run grep -P 'ERROR|WARN' ` and that seems to make the work, supporting perl regexps and ending ok.

```
bats tests/1-check_upgrade_savepoints.bats 
 ✓ check_upgrade_savepoints/check_upgrade_savepoints.sh: v3.1.1 savepoints verified
 ✓ check_upgrade_savepoints/check_upgrade_savepoints.sh: blank upgrade file
 ✓ check_upgrade_savepoints/check_upgrade_savepoints.sh: repeated savepoint
 ✓ check_upgrade_savepoints/check_upgrade_savepoints.sh: no return statement
 ✓ check_upgrade_savepoints/check_upgrade_savepoints.sh: multiple return statements
 ✓ check_upgrade_savepoints/check_upgrade_savepoints.sh: if statement without savepoint
 ✓ check_upgrade_savepoints/check_upgrade_savepoints.sh: wrong savepoint version
 ✓ check_upgrade_savepoints/check_upgrade_savepoints.sh: out of order savepoint steps

8 tests, 0 failures
```

mrmark:
Thanks Eloy, pushed that change, let's see if that fixes the build.  I was kind of fumbling my way through the tests heh.

stronk7:
Are you sure you don't need the `-P` (perl regexp). I using here GNU 3.1 grep in the Mac, and also have looked in a linux box (GNU 2.3 grep) and both needed it to accept pipes.

stronk7:
Side note. Tests are really easy. you just need [bats|https://github.com/sstephenson/bats] somewhere. And then just run, for example (it uses standard shell expansion, afaik in case you want multiple):
```
/path/to/batsrepo/bin/bats tests/1-check_upgrade_savepoints.bats
```

mrmark:
The test wont run without `LOCAL_CI_TESTS_CACHEDIR` and `LOCAL_CI_TESTS_GITDIR` and maybe others, didn't make it that far.  Added `-P` - let's see what Travis has to say.

stronk7:
It seems that travis accepted the `-P` run:

https://travis-ci.org/moodlehq/moodle-local_ci/builds/281353037

(apart from other failures - that I'll try to fix asap - it seems that all the `check_upgrade_savepoints/check_upgrade_savepoints.sh`tests are passing so we can consider this passed.

stronk7:
Broken, thanks!

stronk7:
👅

mrmark:
LOL

PR #182, opened 2017-09-26
Ensure a plugin does not upgrade past its defined version

mrmark:
This checks to make sure that the versions in the upgrade path are not higher than the version defined in the plugin's version.php file.  If this ever happens, it causes a downgrade error.


stronk7:
Maybe you'd consider adding this in order to get the new check covered: 2a78a48

mrmark:
Thanks Eloy!

PR #183, opened 2017-10-20
MDLSITE-5244 rebase_security: Move prepare_npm_stuff to conflict resolution time

andrewnicols:
This needs to happen during conflict resolution, not before.

Different dependency versions can exist at different commits.


stronk7:
Thanks, @andrewnicols . Note @ [MDLSITE-5184](https://tracker.moodle.org/browse/MDLSITE-5184) we are working towards getting rid of current caches and friends, because since 8.2.0 there are some problems working with aliased node_modules...

PR #184, opened 2017-10-30
MDLSITE-5184

stronk7:
Not much to say, look to https://tracker.moodle.org/browse/MDLSITE-5184

This should make things better for next npm "messes". Now stuff is installed on each execution and we are not using own caching of aliased node_modules directories anymore.

There are some tests failing, but they are mustache ones and unrelated with this issue. Will look elsewhere.

Ciao :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ addons:
 php:
     - '7.0'
 env:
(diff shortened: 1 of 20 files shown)


stronk7:
Thanks!

PR #185, opened 2018-01-09
Make local_ci self tests to pass

stronk7:
Along the last 3 weeks I've been slowly moving local_ci tests to put them back to pass. They were failing since some months ago because of various reasons here and there. The changes include:

- Move a good number of tests to more recent branches, reworking the fixture branches and smurf expectations. Surely preparing everything to the deprecation of old tools, now run by grunt...
- Fix various problems, language strings, move to php70, tidy travis file, annoying java problem...

Anyway, just to get everything passing, so we can start performing some changes here and there backed by tests...

Ciao :-)


dmonllao:
Thanks Eloy.

stronk7:
TY, David! Yay!

PR #186, opened 2018-02-07
NOBUG: Grunt to ignore changes caused by 3rd part libs

stronk7:
Some plugins can come with their own 3rd part libs (thirdpartylibs.xml).
That leads to grunt feeding .eslintignore and .stylelintignore to
make js & css linters to ignore those 3rd part libe.

But those changes, make the grunt check to fail as far as they lead
to change in core detected when the check is being performed for a
non-core plugin. So we just proceed to ignore changes in those files
when it's the case.

### DIFF
--- a/grunt_process/grunt_process.sh
+++ b/grunt_process/grunt_process.sh
@@ -4,6 +4,7 @@
 # $gitbranch: Branch we are going to install the DB
 # $npmcmd: Optional, path to the npm executable (global)
 # $npminstall: (optional), if set the script will install nodejs stuff. Else, nodejs managing is external.
+# $isplugin: (optional), if set we are examining a plugin, some exceptions may be applied.
 
 # Let's be strict. Any problem leads to failure.
 set -e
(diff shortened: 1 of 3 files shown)


stronk7:
Before the patch, 3rd part plugin causes core grunt to fail, because it detects changes in ignore files:

http://ci.stronk7.com/view/prehecker/job/Precheck%20remote%20branch/1932/artifact/work/smurf.html
(grunt detail: [link](http://ci.stronk7.com/view/prehecker/job/Precheck%20remote%20branch/1932/artifact/work/grunt.txt))

After the patch, ignore files are explicitly excluded when the checker runs for a plugin:

http://ci.stronk7.com/view/prehecker/job/Precheck%20remote%20branch/1935/artifact/work/smurf.html
(grunt detail: [link](http://ci.stronk7.com/view/prehecker/job/Precheck%20remote%20branch/1935/artifact/work/grunt.txt))

Covered with unit test. Ciao :-)

dmonllao:
Thanks Eloy, merged.

PR #187, opened 2018-03-16
NOBUG: illegal_whitespace: Quote bash vars

andrewnicols:


andrewnicols:
Added new tests and a related fix.

illegal_whitespace/illegal_whitespace.sh
```diff
@@ -27,7 +27,8 @@ countfile=$WORKSPACE/illegal_whitespace_counters_$gitbranch.csv
 mincountfile=$WORKSPACE/illegal_whitespace_mincounter_$gitbranch.csv
 
 # Co to proper gitdir and gitpath
-cd $gitdir && git reset --hard $gitbranch
+cd $gitdir
```
stronk7: Doesn't this need quoting? Like the original patch was doing with all the rest of (worspace?) occurrences? In fact the test is failing because of the missing quotes, isn't it?

stronk7:
Apart from the comment above... have you killed the rest of quotes (WORKSPACE)? Cannot see them anymore.

andrewnicols:
Ah blast - thanks. I reverted the original changes for some testing.

Pushed again,

stronk7:
Thanks!

PR #193, opened 2019-01-15
Bump to PHPCS 2.9.2

stronk7:
To use https://github.com/moodlehq/moodle-local_codechecker/pull/51

(once it has been merged, not before, of course!) ;-)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
     "require": {
-        "squizlabs/php_codesniffer": "2.7.0",
+        "squizlabs/php_codesniffer": "2.9.2",
         "Mustache/mustache": "^2.11"
     }
 }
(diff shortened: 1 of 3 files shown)


stronk7:
Grrr, there are still randoms happening... look to:

https://travis-ci.org/moodlehq/moodle-local_ci/jobs/479949329

Will look after this if there are more place where we should be applying the ordering...

stronk7:
Ok, sorting amended to use some better attributes (initial line, final line, rule, weight...)... let's see.

andrewnicols:
Looks good apart from the Travis fail. Merging anyway since the travis fail looks unrelated.

stronk7:
> Looks good apart from the Travis fail. Merging anyway since the travis fail looks unrelated.

Yeah, it started failing (java installation) recently... let's see if it's consistent and we need to move to openjkd or so... or just a temporal glitch... tests are passing locally (but a couple of remaining fixtures to fix for the new ordering).

Thanks!

PR #195, opened 2019-02-21
 mustache_lint: Wrap stray tags into valid parent.

kabalin:
This fix is partially addressing [MDL-56504](https://tracker.moodle.org/browse/MDL-56504) by adding an extra wrappers to tags that have strict parent requirement according to the spec (e.g. it is invalid to use `<li>...</li>` without wrapping it `<ul>...</ul>` tag).

Bats run output:
If only test (ac593e9) is applied:
```
ci:mustache-lint-partials-stray-tags-fix> bats tests/1-mustache_lint.bats 
 ✓ mustache_lint: Good mustache file
 ✓ mustache_lint: No example content
 ✓ mustache_lint: Example content invalid json
 ✓ mustache_lint: Mustache syntax error
 ✓ mustache_lint: HTML validation issue
 ✓ mustache_lint: Partials are loaded
 ✗ mustache_lint: Partial contains elements with strict parent requirement
   (from function `assert_success' in file tests/libs/bats-assert/src/assert.bash, line 114,
    in test file tests/1-mustache_lint.bats, line 111)
     `assert_success' failed
   
   -- command failed --
   status : 1
   output (5 lines):
     Validating using https://html5.validator.nu
     Running mustache lint from f851201f58662264ad7864dbbf8e2240ce7bf84a to de7907ebe5594bac7cea77c2bf847ea7c4554999:
     /home/ruslan/git/ci/moodle/lib/templates/linting.mustache - WARNING: HTML Validation error, line 2: Stray start tag “td”. (ad><body><td>Hello )
     /home/ruslan/git/ci/moodle/lib/templates/linting.mustache - WARNING: HTML Validation error, line 2: Stray end tag “td”. (ello World</td></bo)
     Mustache lint problems found
   --
   
 ✓ mustache_lint: Full HTML page doesn't get embeded in <html> body
 ✓ mustache_lint: Theme templates load theme partials
 ✓ mustache_lint: Test quote and uniq helpers are working
 ✓ mustache_lint: Test eslint doesn't run when not present
 ✓ mustache_lint: Test eslint runs when npm installed
```
when feature patch (8523227) is applied:
```
ci:mustache-lint-partials-stray-tags-fix> bats tests/1-mustache_lint.bats 
 ✓ mustache_lint: Good mustache file
 ✓ mustache_lint: No example content
 ✓ mustache_lint: Example content invalid json
 ✓ mustache_lint: Mustache syntax error
 ✓ mustache_lint: HTML validation issue
 ✓ mustache_lint: Partials are loaded
 ✓ mustache_lint: Partial contains elements with strict parent requirement
 ✓ mustache_lint: Full HTML page doesn't get embeded in <html> body
 ✓ mustache_lint: Theme templates load theme partials
 ✓ mustache_lint: Test quote and uniq helpers are working
 ✓ mustache_lint: Test eslint doesn't run when not present
 ✓ mustache_lint: Test eslint runs when npm installed

12 tests, 0 failures
```

### DIFF
--- a/mustache_lint/mustache_lint.php
+++ b/mustache_lint/mustache_lint.php
@@ -192,16 +192,59 @@ function print_message($severity, $mesage) {
 }
 
 /**
- * Wrap the template content in a html5 wrapper and validate it
+ * Fix partials and wrap the template content in a html5 wrapper.
+ *
+ * There could be the case when partial template got elements that have a strict
+ * parent element requirement. We can wrap most obvious elements into correct parent,
+ * assuming the partial template is "balanced" i.e. contain equal number of start
+ * and end tags (though, if it is not equial, the partial template will fail
+ * validation anyway). This partial fixing wrapper will not be required if we
+ * find a way to exclude partials from linting at some point (MDL-56504).
+ *
+ * @param string $content the raw template as string.
+ * @return string $wrappedcontent
+ */
+function wrap_content($content) {
+    // Primitive detection if we have full html body, if not, we need to wrap it.
+    if (strpos($content, '<head>')) {
+        return $content;
+    }
+
+    // Mapping of child elements to their required parent elements.
+    $reqparent = [
+        // List.
+        'li' => 'ul',
+        // Table.
+        'td' => 'tr',
+        'th' => 'tr',
+        'tr' => 'table',
+        'thead' => 'table',
+        'tbody' => 'table',
+        'tfoot' => 'table',
+        // Definition list.
+        'dd' => 'dl',
+        'dt' => 'dl',
+    ];
+
+    // Determine if the first element has strict parent element requirement and wrap it if necessary.
+    // Iterate through, as further wrapping may be needed.
+    while (preg_match('@^<(' . join('|', array_keys($reqparent)) . ')[^>]*>@is', trim($content), $matches)) {
+        $parent = $reqparent[$matches[1]];
+        $content = "<{$parent}>\n{$content}\n</{$parent}>";
+    }
+
+    // Wrap in html5 body.
+    $wrappedcontent = "<!DOCTYPE html><head><title>Validate</title></head><body>\n{$content}\n</body></html>";
+    return $wrappedcontent;
+}
+
+/**
+ * Validate template content.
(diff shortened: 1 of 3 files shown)


kabalin:
Thinking if it needs some watchdog for while loops count? (if someone add a new elements mapping that would point to each other, this may produce an infinite loop) :/

danpoltawski:
I like the concept of this solution to avoid more false positives (the the primitive html wrapping was always a bit of a ugly hack).

Though worth noting that a proper declarative metadata type solution might have some value and was how I intended to get this fixed: https://tracker.moodle.org/browse/MDL-56504 It would help the template library too

kabalin:
Yes, to be honest Dan I found MDL-56504 ticket when the code for this solution was ready, thus decided to create PR anyway. Need to make it a habit to search tickets before starting writing anything 😄

stronk7:
Hi,

sorry for the delay... somehow I got this by email but ended well down in my lists!

Just thinking if, maybe... both solutions could coexist. In this way:

1. Nearly sure (contradict me if I'm wrong!)... any of the cases detected by the new {{wrap_content()}} function... are partials, correct? Or maybe they don't need to be? In any case, this is secondary, lets continue...
2. We also can make a pass over all the .mustache files looking for (````{{>````) included partials, yes?
3. When we know that something is a partial, because of 2):
    - If it matches 1), we still can HTML validate it.
    - If it doesn't matches 2), we skip HTML validation (noticing it).

4. If @ [MDL-56504](https://tracker.moodle.org/browse/MDL-56504) there is someday an agreement about how to tag partials... we can start cross-checking those tags with the information we have about partials (detect cases missing the tag... or the opposite, tagged templates that are perfectly valid as main ones as they are.

Not sure about the exact way to put everything working together, just drafting how they may work (the clever wrapping, the detected partials used, the html validation and the future partials tag).

brendanheywood:
So other than a rebase and dust the cobwebs off, whats left to get this fixed?

stronk7:
Cross-linking with https://tracker.moodle.org/browse/MDLSITE-6744, for the records.

PR #196, opened 2019-03-22
MDLSITE-5697 local_ci: accept the short array syntax in dependencies

stronk7:
Backed with unit test covering 1 and multiline cases


junpataleta:
Thanks, Eloy. Merged!

PR #197, opened 2019-04-09
MDLSITE-5716 Fix parsing component names in the commit AMOScript

mudrd8mz:
Components with a digit in the name (such as auth_oauth2) were not
correctly recognised before. This has been fixed upstream in AMOS and
the same fix is being copied over here.

Extending the tests to cover both newly covered cases - valid component
names with a digit, and invalid component name with double underscore.


mudrd8mz:
The upstream fix in AMOS - https://github.com/moodlehq/moodle-local_amos/commit/a9ccb9371df83fd950fb98abc0de696b1f5970c1

stronk7:
Thanks David, this will be auto-deployed in few mins!

PR #198, opened 2019-05-01
 MDLSITE-5739 mustache_lint: error when running on openjdk11

kabalin:
More details:
* https://tracker.moodle.org/browse/MDLSITE-5739
* https://github.com/blackboard-open-source/moodle-plugin-ci/issues/91

As I can see, repo does not include `package-lock.json`, I guess this is intentional, as previously requirement was "^17.3.0" (any compatible version)? I do not use "^18.11.5" as it seems bring up unstable releases above that version when npm is initialised or updated. Perhaps we need to introduce `package-lock.json` to make sure that end user is using the same version which has been integration-tested here?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+dist: xenial
+
 language: php
 
 sudo: false
@@ -6,7 +8,6 @@ addons:
   apt:
     packages:
       - bc
-      - openjdk-8-jre-headless
 
 cache:
   directories:
@@ -18,19 +19,10 @@ services:
 php:
     - 7.1
 
-matrix:
-  include:
-    - dist: xenial
-      env: TEST_SUITE="0-*" DIST="xenial"
-    - dist: xenial
-      env: TEST_SUITE="1-*" DIST="xenial"
-    - dist: xenial
-      env: TEST_SUITE="2-*" DIST="xenial"
-
 env:
-    - TEST_SUITE="0-*" DIST="trusty"
-    - TEST_SUITE="1-*" DIST="trusty"
-    - TEST_SUITE="2-*" DIST="trusty"
+    - TEST_SUITE="0-*"
+    - TEST_SUITE="1-*"
+    - TEST_SUITE="2-*"
 
 install:
   - git clone --depth 1 https://github.com/sstephenson/bats.git $HOME/bats
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,15 @@ cache:
 php:
     - 7.1
 
+matrix:
+  include:
+    - dist: xenial
+      env: TEST_SUITE="0-*"
+    - dist: xenial
+      env: TEST_SUITE="1-*"
+    - dist: xenial
+      env: TEST_SUITE="2-*"
+
 env:
     - TEST_SUITE="0-*"
     - TEST_SUITE="1-*"
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,22 +12,25 @@ cache:
   directories:
       - $HOME/cachedir
 
+services:
+  - mysql
+
 php:
     - 7.1
 
 matrix:
   include:
     - dist: xenial
-      env: TEST_SUITE="0-*"
+      env: TEST_SUITE="0-*" DIST="xenial"
     - dist: xenial
-      env: TEST_SUITE="1-*"
+      env: TEST_SUITE="1-*" DIST="xenial"
     - dist: xenial
-      env: TEST_SUITE="2-*"
+      env: TEST_SUITE="2-*" DIST="xenial"
 
 env:
-    - TEST_SUITE="0-*"
-    - TEST_SUITE="1-*"
-    - TEST_SUITE="2-*"
+    - TEST_SUITE="0-*" DIST="trusty"
+    - TEST_SUITE="1-*" DIST="trusty"
+    - TEST_SUITE="2-*" DIST="trusty"
 
 install:
   - git clone --depth 1 https://github.com/sstephenson/bats.git $HOME/bats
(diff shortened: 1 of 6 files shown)


kabalin:
I think travis fails on https://github.com/moodlehq/moodle-local_ci/blob/master/mustache_lint/mustache_lint.sh#L30 (e.g. "11.0.2" > "1.8" 😉 ), will see tomorrow if it is possbile to do proper semver comparison.

mustache_lint/mustache_lint.sh
```diff
@@ -24,10 +24,12 @@ mydir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
...
 then
-    if java -version 2>&1 >/dev/null | grep -q "java version"
+    if java -version 2>&1 >/dev/null | grep -Eq "(java|openjdk) version"
     then
-        javaversion=$(java -version 2>&1 | awk -F '"' '/version/ {print $2}')
-        if [[ "$javaversion" > "1.8" ]]
+        # Extract major and minor digits of version and compare it as float
+        # numbers. Not error proof, but sufficient to ensure it is >= 1.8
+        javaversion=$(java -version 2>&1 | awk -F '"' '/version/ {print $2}' | sed 's/^\([0-9]*\.[0-9]*\).*$/\1/')
+        if [ $( echo "${javaversion} >= 1.8" | bc ) == 1 ]
```
stronk7: Won't this fail the day enter into the 2 digit comparisons?

```echo "1.10 >= 1.8" | bc``` = 0

File: .travis.yml [hunk 2]
```diff
@@ -12,13 +12,25 @@ cache:
   directories:
       - $HOME/cachedir
 
+services:
+  - mysql
+
 php:
     - 7.1
 
+matrix:
```
stronk7: Any reason we need this new section included? Cannot the 3 new xenial runs go to the ```env``` one and forget? I personally find things more readable if together. Not critical though.

package.json
```diff
@@ -4,6 +4,6 @@
   "description": "Tools for CI with Moodle",
   "private": true,
   "dependencies": {
-    "vnu-jar": "^17.3.0"
+    "vnu-jar": "18.11.5"
```
stronk7: I think it's ok if we start providing a lock file, as you comment in the issue. Cannot imagine any downside, but better control and safer results. My +1 here.

.travis.yml
```diff
@@ -12,13 +12,25 @@ cache:
   directories:
       - $HOME/cachedir
 
+services:
```
stronk7: Don't oppose but isn't always included?

https://docs.travis-ci.com/user/database-setup/#mysql

mustache_lint/mustache_lint.sh
kabalin: Yep, that is why I state it is sufficient, but not error proof :) There is no java 1.10 in such  notation (to the best of my knowledge). Java 10 in Oracle versioning was [18.3](https://www.jcp.org/en/jsr/detail?id=383) (they switched to year/month based versioning since version 10), I can't check how that looks in installation, but if openjdk package is used, it reads 11.x (there are no [Debian](https://packages.debian.org/search?keywords=openjdk)/Ubunitu packages for anything between 1.8  and 11). So, I think for purpose of ensuring that version is higher than 1.8, the suggested comparison is sufficient (at least until we find an edge case where this does not work 😄 )

.travis.yml
kabalin: Not in Xenial https://docs.travis-ci.com/user/reference/xenial/#services-disabled-by-default

.travis.yml
kabalin: No, you can't change disto using env definition, only option that should work in theory is (with leaving env as is):
```
dist:
 - trusty
 - xenial
```
But it does not work that way, so had to use matrix instead.
(that DIST var is dummy, it is only to see which distro was used when you are on Travis build page)

mustache_lint/mustache_lint.sh
kabalin: Alternatively, it is possible to compare that version is not lower than 1.8, though not sure if that will make it more error proof. What do you think?

package.json
kabalin: I added commit with lock file.

kabalin:
Thinking more. May be we should group all mustache tests and run them in all available java environments, and not differentiate between trusty and xenial?

mustache_lint/mustache_lint.sh
stronk7: Nah, all ok. If some day there is a situation when it fails... we can amend it, no big problem.

.travis.yml
stronk7: yeah, had to add it recently to a number of plugins too.

package.json
stronk7: super!

stronk7:
Thanks @kabalin !

Just a couple of questions:

1) With xenial becoming the default one... maybe we'll need to also add explicitly the trusty runs to the matrix/include? If I understand it properly we may end running xenial twice unless we explicitly set the dist for each run, isn't it?

2) About testing on all available java environments I really don't know if it's worth it* . Surely I'd wait somebody to report that it's not working properly with java xxxxx instead of continuously try to keep the tests on top of a bunch of flavors and versions.

* note that, personally, i'm not sure, either, if it's worth testing it over multiple ubuntu distributions, it just doubles the running time when, by default, I'd expect things to continue working the same. But I don't oppose, just 1) above called my attention.

Other than that 1), I'm happy with this one as is, tell me that do you think so we can advance it...

Ciao :-)

stronk7:
Any thought about that 1) above, @kabalin ? I had this completely forgotten, sorry, just arrived here today by coincidence (another PR).

kabalin:
Hi @stronk7, sorry I missed the your first comment completely (I was on holidays most of June).
1) - you are right :)
2) - I actually was thinking thought the same before reading "*" part. We know it works fine with openjdk8, so testing explicitly if it "still works" with old version probably not required. The default recommendation for anyone who has issues in older distro would be just to switch to xenial, which we tested against in this CI scenario. To the best of my knowledge, moodle-plugin-ci that is using this package as dependency has no critical issues using xenial either, apart of mustache linting, which this fix is addressing.

Changes will follow in a moment.

stronk7:
Here we go, thanks!

kabalin:
👍

PR #200, opened 2019-07-22
Remove bogus/outdated link to tracker issue

stronk7:
And add some basic general links about running php jobs with
different CI/containerised alternatives.


stronk7:
(self-merged :-) )

PR #201, opened 2019-09-03
MDLSITE-5931 local_ci: Be ready for a move to package-lock.json

stronk7:
Just in case we move at some point, now local_ci will work with
both package-lock.json and shrinkwrap.json (after all, all it needs
to do is to ignore both when looking for changes after running grunt)


andrewnicols:
I'm actually wondering why we need to do this at all. We should now _never_ see changes. If we do, then they a problem.

If people are using the correct NPM version (as dictated by the new nvmrc), and the node_modules does not contain any incorrect modules, then it will never change.

stronk7:
Yeah, this needs a little bit of history...

1) since we started to control (and document) the desired nodejs/npm versions to use... they have broken the lock files format a couple of times. And that was causing all the travis and CI grunt tests to fail, because differences in the lock file were found (a simple ```git ls-files -m``` is performed to check if grunt has performed any change).

2) as far as that was extremely noisy  (all reds everywhere) and had happened twice, we decided to, explicitly, exclude the lock file (that may change). And delegate the task of continuously looking for future npm breaks to [MDLSITE-5536](https://tracker.moodle.org/browse/MDLSITE-5536).

That way a) we'll be checking for breakages before they happen and b) problems won't affect to grunt jobs that will check that everything else (built js, css, ignore files)... are ok being immune to the npm problem.

So this job just makes the grunt job immune to both lock files, in case we switch to the newer one at some point in the future and they break (format changes) again. Just that, exclude the lock files from any grunt-changes consideration, to isolate the job from any npm problem.

Ciao :-)

PS: Note that we haven't implemented [MDLSITE-5536](https://tracker.moodle.org/browse/MDLSITE-5536) yet, but we really should, to keep it running the latest LTS version and warn us when an update is wroth it (or the lock format has changed again). In practice, it's not critical, because all us do run npm install daily and immediately would notice any change in the lock files, but for sure it's a wannabe to have it detected and warning us.

andrewnicols:
But if we switch to this nvmrc as proposed we will be using a fixed version of node and should not be affected by them

stronk7:
If people uses nvm, yes (I do since ever).

In any case, as said, the idea was to separate/isolate failures so grunt-changes results weren't affected by an unrelated, npm problem. That's what the patch provides (moodle's travis does pretty much the same, btw, explicitly excluding the shrinkwrap file).

stronk7:
pong?

stronk7:
I'm closing this, outdated, now everybody uses nvm. And we continue using `npm-shrinkwrap.json`

PR #204, opened 2019-09-29
Launch predefined jobs against issues returned by a query

stronk7:
This job just allows you to perform JQL queries against
a Jira server, launching a number of defined Jenkins jobs for each
issue found and performins some post-actions for each issue.

We are going to use it with a criteria named "awaiting_integration"
so, every user that lands to integration gets some CI jobs executed
for every branch in the issue.

At Jenkins level we'll be throttling those (DEV) jobs to prevent
this script to end using all the available workers.

It's important to make the schedulemins parameter to match the
intervals (in minutes) in the job schedule plan (cron) or some
issues could be processed twice or skipped.

Finally note that the current criteria includes a extra condition:

AND participants not in (tobic)

So we'll be processing an issue only once. This can be tidied/chaged
later, but for now it's a basic protection trying to avoid an issue
going back and forth to consume too much infrastructure resources.

Also note that, for now, we have disabled the Oracle PHPUnit job, just
to sabe some workers until we have the job controlled and decide that
we can add more jobs to it.

This job will be launched by the "TR - Bulk pre-launch DEV jobs
(awaiting_integration)" job that has been already created and is
waiting for this to land to perform the final configuration and
schedule it.

### DIFF
--- a/tracker_automations/bulk_prelaunch_jobs/bulk_prelaunch_jobs.sh
+++ b/tracker_automations/bulk_prelaunch_jobs/bulk_prelaunch_jobs.sh
@@ -1,0 +1,137 @@
+#!/usr/bin/env bash
+# Look all reopened issues under current integration and move them out
+#jiraclicmd: full execution path of the jira cli
+#jiraserver: jira server url we are going to connect to
+#jirauser: user that will perform the execution
+#jirapass: password of the user
+#jenkinsserver: Full URL of the jenkins server where the jobs will be launched.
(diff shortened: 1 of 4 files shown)


stronk7:
Going ahead with this, so we can test it under CI:

https://ci.moodle.org/view/Tracker/job/TR%20-%20Bulk%20pre-launch%20DEV%20jobs%20(awaiting_integration)/

Have run a number of time here from terminal and everything is looking acceptably ok.

Ciao :-)

PR #206, opened 2020-01-07
Ignore ics files in illegal whitespace checks

andrewnicols:
ICS files must have a proper line break so will always fail the illegal whitespace check. Let's just ignore them. They're a false positive.

### DIFF
--- a/define_excluded/define_excluded.sh
+++ b/define_excluded/define_excluded.sh
@@ -86,6 +86,7 @@ yui/build/
 *.csv
 *.gif
 *.jpg
+*.ics
 *.png
 *.svg"
 


stronk7:
Just guessing if it wouldn't be better to follow the trend and, simply, add the `*.ics` type here:

https://github.com/moodlehq/moodle-local_ci/blob/master/define_excluded/define_excluded.sh#L90

I cannot imagine any check that should, ever, consider those files, so maybe it's better to apply for it there. For your consideration.

Ciao :-)

PS: About travis failures, running the database comparator... it's normal now that master requires php 7.2. I'm working on a fix for that....

sarjona:
@andrewnicols I agree with Eloy and, in that particular case, I think it would be better to exclude the *.ics files. Or is there any special benefit checking those files?

sarjona:
Thanks Andrew!! :-)

However, won't be better to add the *.ics in a new line into the excluded variable: https://github.com/moodlehq/moodle-local_ci/blob/master/define_excluded/define_excluded.sh#L9?

PR #208, opened 2020-04-12
MDLSITE-6103 local_ci: gherkinlint json to checkstyle

stronk7:
Ensure that checkstyle_converter.php is able to
properly convert from gherkinlint JSON output format
to checkstyle.

PR #209, opened 2020-04-13
Cibot gherkinlink

stronk7:
This is built on top of #208, so better considered after it. Basically adds the new individual gherkinlint check, modifies tests to pass with it incorporated to out put and then, in separate commit, adds own tests.


andrewnicols:
Looks good to me. Are we now in place to merge this or do we need to wait for any core issues first?

stronk7:
I don't think we are waiting for anything (sounds to me that you fixed the current 2-blank spaces case in some MDL issue)...

Also note that surely you can just merge this and close #208 because this is built on top of that.

 let me check... if we are clean right now (although it won't make much difference if there are current errors, because it only reports lines belonging to the patch). One min...

stronk7:
Confirmed, have run gherkinlint for 35, 36, 37, 38 and master (moodle.git) and it ended ok

```
$ git grunt gherkinlint && echo "yay!"
Running "gherkinlint" task

Done.
yay!
```

PR #213, opened 2020-05-12
Fix change of behavior in eslint/stylelint (exit when no files)

stronk7:
Old eslint versions, when there weren't any JS file to inspect
just exited silently and without error (exit status = 0).

But since eslint v6.8.0, if there aren't JS files, it exits
with error (exit status = 1). A new option,
--no-error-on-unmatched-pattern, has been created to provide
the BC bahavior.

So we need to discern which eslint is being used. Initally I
did it looking for --version... but it's inexact.

So, surely, the safest way is to verify if the new option exists
or no (using --help). And decide based on that.

That's exactly what this patch does. I've also added a TODO
because one everybody is already using nodejs 14 we won't need
this hack anymore.

Also, added pretty much the same for stylelint, but using the
--allow-empty-input option.

Finally note that there is another eslint use (within the mustache
linter) that is not fixed here. It will come in a new PR because
the fix is different (eslint is executed from within PHP).

### DIFF
--- a/remote_branch_checker/remote_branch_checker.sh
+++ b/remote_branch_checker/remote_branch_checker.sh
@@ -416,7 +416,15 @@ if [ -f $WORKSPACE/.eslintrc ]; then
     echo "Info: Running eslint..."
     eslintcmd="$(${npmcmd} bin)"/eslint
     if [ -x $eslintcmd ]; then
-        $eslintcmd -f checkstyle $WORKSPACE > "${WORKSPACE}/work/eslint.xml"
+        # TODO: Remove this once everybody is using nodejs 14 or up.
+        # We need to invoke eslint differently depending of the installed version.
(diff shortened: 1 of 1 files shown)


stronk7:
Have run the patch above directly @ integration.moodle.org and it seems now those JS and CSS less patches are passing:

https://integration.moodle.org/job/Precheck%20remote%20branch/86717/

Also, ignore travis execution. Maybe will fail, not sure. I've another ongoing issue to fix all tests together.

stronk7:
> Finally note that there is another eslint use (within the mustache
linter) that is not fixed here. It will come in a new PR because
the fix is different (eslint is executed from within PHP).

Just for the records, no action is needed for the `eslint` use within the mustache linter. It validates JS inline (extracted from the templates, not files). So the problem cannot be reproduced there, always there will be an input to process.

Ciao :-)

PR #214, opened 2020-05-25
Continuous manage queues (1 of 2)

stronk7:
This script adds some automatisms helping to manage the integration queues:
 - candidates queue: issues awaiting from integration not yet in current.
 - current queue: issues under current integration.

The automatisms are as follow:
 1) Move "important" issues from candidates to current.
 2) Keep the current queue fed with issues when it's under a threshold.

The criteria to consider an issue "important" are:
 1) It must be in the candidates queue, awating for integration.
 2) It must not have the integration_held or security_held labels.
 3) It must not have the "agreed_to_be_after_release" text in a comment.
 4) At least one of this is true:
   a) The issue has a must-fix version.
   b) The issue has the mdlqa label.
   c) The issue priority is critical or higher.
   d) The issue is flagged as security issue.
   e) The issue belongs to some of these components:
     - Privacy
     - Automated functional tests (behat)
     - Unit tests

This job must be enable only since freeze until release day.

Note this implements only the 1st automatism, another commit will add
the 2nd one.

### DIFF
--- a/tracker_automations/continuous_manage_queues/continuous_manage_queues.sh
+++ b/tracker_automations/continuous_manage_queues/continuous_manage_queues.sh
@@ -63,6 +63,8 @@ movemax=${movemax:-3}
 # the search and the update in order to log all the closed issues within jenkins ($logfile)
 
 # 1) Move "important" issues from candidates to current.
+
+# Get the list of issues.
 ${basereq} --action getIssueList \
            --search "filter=14000
                  AND NOT filter = 21366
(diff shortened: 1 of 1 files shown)


stronk7:
Has been tested with MDL-68750 and some other issue. Seems to be working ok. Right now there aren't important issues to be moved.

TODO:
1. Create the TR job to run this.
2. Implemente the 2nd automatism.

When:
Tomorrow, ;-)

stronk7:
And 2nd automatism implemented in new commit. You can see it in action in issues like MDL-52138 (we were below 6 issues in current, so up to 3 issues where added - both numbers are configurable).

So, updated TODO:

1. Merge this if acceptable.
2. Create the TR job to run this.
3. Document in the process when this job has to be enabled (immediately after freeze) and disabled (when packaging the major).
4. Party.

Final niao! :-)

sarjona:
Hi Eloy!
Thanks for working on this! The patch looks good and it's really well documented and easy to follow! :-) Good job!!

stronk7:
Thanks @sarjona.

I've completed the TODO above (have added links), so this should be considered done.

I'm checking what happens with issues (non-bug ones) that haven't been held yet... I suspect that, with current query... they will be moved to current integration automatically.

So, surely I'll self-merge a tiny change to the query to ensure that the job only handles Bug issues, leaving new features/improvements apart.. let's see.

PR #215, opened 2020-05-31
Add both sqlsrv phpunit and @app (experimental) jobs to ToBiC

stronk7:


stronk7:
Note I've reverted the \@app experimental job, it's failing badly, also @ CIs, so better don't add it till stable.

PR #221, opened 2020-09-07
Add possibility to exclude selected Mustache files from HTML validation

mudrd8mz:
When developing for the Moodle Mobile App, the Mustache templates may
contain Ionic3 / Angular specific syntax that itself is not valid HTML.
The patch adds an option to skip the HTML validation part of the
Mustache linting.
 
To ignore HTML validation errors, create a file `.mustachelintignore` in
the directory with Mustache templates and define the file name patterns
for the files to be excluded from HTML validation. PHP function
`fnmatch()` is used to search for matches. Lines starting with the hash
sign are considered as comments.

### DIFF
--- a/mustache_lint/mustache_lint.php
+++ b/mustache_lint/mustache_lint.php
@@ -73,6 +73,11 @@ if (strpos($FILENAME, $CFG->dirroot) !== 0) {
     exit(1);
 }
 
+// Ignore the .mustachelintignore file itself.
+if (basename($FILENAME) === '.mustachelintignore') {
+    exit(0);
+}
+
 require_once(__DIR__.'/simple_core_component_mustache_loader.php');
 require_once(__DIR__.'/js_helper.php');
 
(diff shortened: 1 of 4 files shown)


stronk7:
Q: Do we want to completely ignore them or just prevent the html validation to happen? If I'm not wrong the problems linked the other day @ dev chat were all them html validation ones.

mudrd8mz:
Ah, right. That's a good point! We could easily just skip the `check_html_validation()` for them, but still run all the other checks. I am happy to amend the patch in that way.

The only thing - I think we should have a better name for the dot file then, as it is not "mustache lint ignored" any more. I am thinking of something like `.htmlvalidationignored` maybe?

mudrd8mz:
Nah. I think that `.mustachelintignore` will work best for the consistency with files like `.eslintignore` or `.gitignore`.

I have amended the patch so that now the mustache files are still checked for all the requirements, they just do not need to produce a valid HTML.

stronk7:
Thanks @mudrd8mz , this has been merged and will be used by CiBoT (remote branch checker) in minutes.

moodle-plugin-ci will start using it on next release.

PR #224, opened 2020-09-24
Ensure that no vendor dir is ever linted

stronk7:
Previously we were excluding vendor directories
but only when "fullmode" (all files) were inspected.

Now this rule also applies to "intercommit" mode (only
files modified between 2 commits).

Covered with tests, checking both 1st level and inner
level vendor directories.


stronk7:
Going ahead myself. Everything is working ok and the problem is affecting some plugins using "vendor" stuff (polyfills and similar are specially problematic for direct lint).

PR #225, opened 2020-10-07
New optional param (jobtype) added. Defaults to all.

stronk7:
Can be used to select which type of job we want to run
(different from all). Useful to sabe some worker slots
when only some job type is failing.

Current aceepted values, apart from "all" are:
- phpunit
- behat-app (it's commented out right now).
- behat-chrome
- behat-goutte


tracker_automations/bulk_prelaunch_jobs/bulk_prelaunch_jobs.sh
```diff
@@ -12,6 +12,7 @@
 #cf_branches: pairs of moodle branch and id for "Pull XXXX Branch" custom field (master:customfield_10111,....)
 #criteria: "awaiting integration"...
 #schedulemins: Frecuency (in minutes) of the schedule (cron) of this job. IMPORTANT to ensure that they match or there will be issues processed more than once or skipped.
+#jobtype: defaulting to "all", allows to just pick one of the available jobs: phpunit, behat-chrome, behat-goutte.
```
scara, line 15: Small typo: `behat-app` missing here.

HTH,
Matteo

stronk7: Hi Matteo, I did not add it because right now it's disabled/commented out, surely could have been added, yes.

PR #226, opened 2020-10-09
Ensure that eslint is called relative to the moodleroot

andrewnicols:
ESLint searches for plugins relative to the CWD, and is unable to find
plugins when called from outside of the moodleroot, which the plugin
supports.

### DIFF
--- a/mustache_lint/js_helper.php
+++ b/mustache_lint/js_helper.php
@@ -64,7 +64,7 @@ class js_helper {
         ];
 
         $cmd = "$eslint --stdin --no-eslintrc --config {$this->moodleroot}/.eslintrc --format=json";
-        $proc = proc_open($cmd, $pipesspec, $pipes);
+        $proc = proc_open($cmd, $pipesspec, $pipes, $this->moodleroot);
         // Send the JS to stdin.
         fwrite($pipes[0], $this->js);
         fclose($pipes[0]);
--- a/tests/1-mustache_lint.bats
+++ b/tests/1-mustache_lint.bats
@@ -179,3 +179,43 @@ setup () {
     assert_output --partial "lib/templates/js_test.mustache - WARNING: ESLint warning [camelcase]: Identifier 'my_message' is not in camel case"
     assert_output --partial "lib/templates/js_test.mustache - WARNING: ESLint warning [no-alert]: Unexpected alert. ( alert(my_message); )"
 }
+
+@test "mustache_lint: Test eslint runs ok when invoked from any directory" {
+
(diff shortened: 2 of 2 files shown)


andrewnicols:
Note: to see this issue in action simply run the mustache_lint from the parent directory of any moodle install:

```
cd path/to/your/moodle
git clone git://github.com/moodlehq/moodle-local_ci local/ci
npm ci
composer install
cd local/ci
composer install
cd -
moodleroot=`pwd`
php "$moodleroot"/local/ci/mustache_lint/mustache_lint.php --filename="$moodleroot/lib/templates/paged_content.mustache"  --basename="$moodleroot" --validator="$moodleroot/node_modules/vnu-jar/build/dist/vnu.jar"
echo "All good for the first run"
cd ..
php "$moodleroot"/local/ci/mustache_lint/mustache_lint.php --filename="$moodleroot/lib/templates/paged_content.mustache"  --basename="$moodleroot" --validator="$moodleroot/node_modules/vnu-jar/build/dist/vnu.jar"
echo "Failed"
```

stronk7:
Have added an extra commit to your branch, @andrewnicols , that covers the explained case, with the check being called while @ moodleroot/local/ci, moodleroot and /tmp/xxxx.

Without the patch the 3rd case fails, with the patch it passes. So all right here. Thanks!

stronk7:
Aha, that commit needs more love, now I see how this is run within travis (completely separate dir). Will try again tomorrow.

stronk7:
It's really unfortunate, it seems that old eslint versions didn't have that requirement, so, if i test this with Moodle 3.1.3 (that is what all the mustache-lint tests do by default), then it passed without the patch.

Only if I bump to current (39_STABLE, for example) I can reproduce the error via tests. Ok, trying to allow both the 31 and 39 tests to co-exist.

I've added a note about to bump all tests to more recent branches (and cleanup old code), because using 31 is no way. But that won't be done today.

Trying...

stronk7:
Ok, with the new test, everything is passing (and failing without @andrewnicols fix). So I'm going to merge this now.

Side note: part of the troubles I was having with the tests is that the {{prepare_npm_stuff/prepare_npm_stuff.sh}} always needs to be sourced. Normally that doesn't cause problems, but in this case, where 2 different nodejs versions were being installed (one for local_ci, another for core), the 2nd one was not setting env vars properly, so the local_ci one was used (and missing stuff).

Once this is running I'll create another PR to change all the uses of {{prepare_npm_stuff/prepare_npm_stuff.sh}} (in tests) to be properly sourced, and will change also code from "." to "source". Clearer, IMO.

Ciao :-)

stronk7:
Merged, thanks!

PR #227, opened 2020-11-17
Bump CiBoT to local_codechecker 3.0.0beta (phpcs3)

stronk7:
Still not in master, we are moving to phpcs3 to get
some good real runs so people can report problems.

Once the final version is released we'll move back to master.


stronk7:
Once this passes, I'll be self-merging it and performing the needed configuration changes @ integration.moodle.org. Then, all new CiBoT runs will start using codechecker 3.0beta.

Let's see how it goes, local tests against real remote branches have worked ok.

PR #228, opened 2021-01-13
Accept return bool type hinting on upgrade functions

stronk7:
New xmldb_upgrade_xxxx() functions should start to come
with return (: bool) type hinting. We need to support that.

So I've added a non-capturing and optional group to the regexp matching the
function line and then everything continues working ok. Backed with test.


sarjona (approved):

sarjona:
Thanks for raising and fixing this, Eloy (I almost got crazy trying to figure out why the script was failing!). The patch looks great and I've been able to check that is working as expected.

stronk7:
Thanks!

PR #229, opened 2021-02-26
Add a new jobtype to launch all behat tests (goutte and chrome) together

stronk7:
In many situations we want to run all behat tests, keeping phpunit apart
(specially because we run sqlsrv by default for better checking
cross-db).

This issue introduces a new behat-all job type that will behave exactly
that way. Trivial patch.

I've already configured [our jobs](https://ci.moodle.org/view/Tracker/job/TR%20-%20Bulk%20pre-launch%20DEV%20jobs%20(list_of_mdls)/) to show the new options there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
-dist: xenial
-
 language: php
 
-sudo: false
+os: linux
+dist: xenial
 
 addons:
   apt:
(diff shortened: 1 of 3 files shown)


stronk7:
Wow, have tried a bunch of alternatives to try to get the `compare-databases` tests random timeout failures down. Think it's sorted now, by moving MySQL to tmpfs like we do in core's Travis, in `moodle-ci-runner`... let's see if it works as expected.

PR #231, opened 2021-03-12
Switch all ToBiC jobs to firefox (failing less right now than Chrome)

stronk7:
Mainly because of the new Chrome 89 has come with some surprises, see:

https://tracker.moodle.org/browse/MDL-71108

But also because lately we were getting way less failures with Firefox
than with Chrome, we are switching all ToBiC jobs criteria to run
with Firefox.


stronk7:
Self-merged, going to give FF a try.

PR #232, opened 2021-03-15
Reset the ismasteronly flag for every issue being checked

stronk7:
Today the first Moodle 4.0 only issue has landed, hence it
has got the first master-only label.

Problem is that such label was not being reset properly in the
loop and once set... it was applied to all subsequent issues.

By moving it to within the loop, we ensure it has the
correct default (empty/false) for every issue (unless
it's detected to be present).


stronk7:
Passed tests, hence merging...

PR #233, opened 2021-03-23
Stop requiring Travis

andrewnicols:
Most people have moved away from Travis so we should stop warning if
it is not used.

THis change modifies the Travis integration tests to mark the test as
SKIPped if the integration is not setup. A warning is still used if
travis-ci.org is configured but not travis-ci.com.

Note: I have not written a Github Actions integration to replace this yet.

### DIFF
--- a/tests/1-travis-branch-checker.bats
+++ b/tests/1-travis-branch-checker.bats
@@ -18,7 +18,7 @@ load libs/shared_setup
     # A branch which Dan has had integrated and won't touch again (and is protected for this purpose). Using travis-ci.org
     ci_run_php travis/check_branch_status.php --repository=git://github.com/danpoltawski/moodle.git --branch=MDL-52127-master
     assert_success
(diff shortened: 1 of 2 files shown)


stronk7:
Ah,

I was sure I had some issue created somewhere to make both gha detections and travis to work better together, so whenever the former exists it doesn't look for the later.

https://tracker.moodle.org/browse/MDLSITE-6355

It's tricky to implement it via JS (same-origin restrictions) in the Tracker, although I've an idea, but easy here, @ local_ci (normal curl request is enough).

In any case, nothing of the above negates this issue, just adding a little bit of context to it.

stronk7 (approved):
Changes are good enough, and they allow to expand later when we add support for GHA status detections (the MDLSITE commented above).

So all right, thanks!

stronk7:
Merging with the eyes closed and the nose covered! :D :-P. Thanks!

PR #234, opened 2021-03-25
Allow to decide the target for the ToBiC comments

stronk7:
Right now, all the comments in the criteria are
harcoded to Integrators.

This change allow you to pick any other role in the project
with blank meaning visible to all.

The jobs @ ci.moodle.org has been already modified to support
this new parameter that will be configurable when any of the
jobs there are executed manually. By default, they continue
restricting to Integrators, until we decide to open them.

Local tests seem to work, but I've not been able to test the Tracker API call itself, so... cross fingers :-)


stronk7:
Thanks!

(as expected I had to add extra commit, removing a couple of double-quotes. For me it's impossible to get it ok at the first time, specially when it's a shell script calling to another one calling to one java).

Anyway, seems to be working now, so yay!

PR #235, opened 2021-05-16
Remove very old less checker and shifter walk

stronk7:
They aren't used since years ago and some of the npm
dependencies aren't available anymore.

Also, amend a setup test, so we don't have to update it
on every codechecker update.


stronk7:
Self-merging as far as this is only removal of old stuff + small amending tests.

PR #236, opened 2021-05-19
First cut of the normal queues manager

stronk7:
As discussed, the idea is to use this "normal periods" manager instead of the current "bulk-move on Mondays and bulk-delay on Thursdays (after cutoff)". Basically what we do in the continuous period, when the corresponding queues manager keeps us fed and does also other things (holding...). But the idea is to make issues "flow" instead of the current bulk way commented above.

If we perform this change still will have to:

1. #237 : see how we send the rebase message to issues already in continuous (because now they aren't going to be moved back to candidates).
2. ✔️ Agreed, like in continuous, also in the [integration exposed entries](https://moodle.org/mod/forum/discuss.php?d=422487#p1702271). Agree how we account status (like we do in continuous, queue sizes, basically) and report it.
3. ✔️ Decided, we don't delay issues (neither send a message) anymore.  Decide if we still send the "cutoff" message or no.
4. ⚠️ They are disabled for now, will remove them soon (once this is working ok). Completely remove the 2 "bulk" jobs ([Monday bulk move](https://ci.moodle.org/view/Tracker/job/TR%20-%20Move%20awaiting%20issues%20to%20current%20integration/) and [Thursday delay](https://ci.moodle.org/view/Tracker/job/TR%20-%20Delay%20awaiting%20issues/)) commented above.
5. ✔️ Done: Amend both [integration docs](https://docs.moodle.org/dev/index.php?title=Integration_Review&type=revision&diff=58891&oldid=58077) and [release process docs](https://docs.moodle.org/dev/index.php?title=Release_process&type=revision&diff=58886&oldid=58847).

----

Way simpler than the continuous one, only has to:

- Move "important" inssues to current.
- Keep current fed with issues when under a threshold.
- Raise the integration priority of issues waiting as candidates too long (configurable number of days).

### DIFF
--- a/tracker_automations/continuous_manage_queues/continuous_manage_queues.sh
+++ b/tracker_automations/continuous_manage_queues/continuous_manage_queues.sh
@@ -29,7 +29,7 @@
 #      - Automated functional tests (behat)
 #      - Unit tests
 #
-# This job must be enabled only since freeze day to packaging day.
+# This job must be enabled only since freeze day to the end of on-sync period, when normal weeklies begin.
 #
 # Parameters:
 #  jiraclicmd: fill execution path of the jira cli
--- a/tracker_automations/normal_manage_queues/lib.sh
+++ b/tracker_automations/normal_manage_queues/lib.sh
@@ -1,0 +1,151 @@
+#!/usr/bin/env bash
+# Functions for normal_manage_queues.sh, look there
+# for details and nomenclature used.
+
+# Let's go strict (exit on error)
+set -e
+
+# Verify everything is set
+required="WORKSPACE jiraclicmd jiraserver jirauser jirapass"
+for var in $required; do
+    if [ -z "${!var}" ]; then
+        echo "Error: ${var} environment variable is not defined. See the script comments."
+        exit 1
+    fi
+done
+
+# A, move "important" issues from candidates to current
+function run_A() {
+    # Get the list of issues.
+    ${basereq} --action getIssueList \
+               --search "filter=14000
+                     AND NOT filter = 21366
+                     AND (
+                       filter = 21363 OR
+                       labels IN (mdlqa) OR
+                       priority IN (Critical, Blocker) OR
+                       level IS NOT EMPTY OR
+                       component IN ('Privacy', 'Automated functional tests (behat)', 'Unit tests')
+                     )" \
+               --file "${resultfile}"
+
+    # Iterate over found issues and perform the actions with them.
+    for issue in $( sed -n 's/^"\(MDL-[0-9]*\)".*/\1/p' "${resultfile}" ); do
+        echo "Processing ${issue}"
+        if [ -n "${dryrun}" ]; then
+            echo "Dry-run: $BUILD_NUMBER $BUILD_TIMESTAMP ${issue} moved to current: important"
+            continue
+        fi
+        # For fields available in the default screen, it's ok to use updateIssue or SetField, but in this case
+        # we are setting some custom fields not available (on purpose) on that screen. So we have created a
+        # global transition, only available to the bots, not transitioning but bringing access to all the fields
+        # via special screen. So we'll ne using that global transition via progressIssue instead.
+        # Also, there is one bug in the 4.4.x series, setting the destination as 0, leading to error in the
+        # execution, so the form was hacked in the browser to store correct -1: https://jira.atlassian.com/browse/JRA-25002
+        # Commented below, it's the "ideal" code. If some day JIRA changes that restriction we could stop using
+        # that non-transitional transition and use normal update.
+        #${basereq} --action updateIssue \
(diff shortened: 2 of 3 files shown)


andrewnicols:
I wonder whether we need a `run_C()` too which moves items to Current when they've been waiting for integration for a period. The risk here is that we lose track of all of the small items because the big items keep the current queue at threshold.

stronk7:
> I wonder whether we need a run_C() too which moves items to Current when they've been waiting for integration for a period. The risk here is that we lose track of all of the small items because the big items keep the current queue at threshold.

Yeah, have been thinking about that... but I have some resistance against exceptionally move too many issues (apart from the "important" ones) to current integration, that may render the "continuous" approach not good (flood vs flow of issues).

Maybe what we could do is add that C point and, instead of moving... just when an issue has been sitting in the candidates queue, awaiting for integration over X time (say 2 weeks, for example... we raise its integration priority). With that, we guarantee that whenever the manager picks issues, they will be picked the very first.

That way we don't get issues waiting forever and, also, we keep the flow... thoughts? Of course, we can try and, once we see how things are going... make further adjustments / rules...

> The only thing I'll mention is that there's a lot of duplication which could be moved to a move_issue() function taking issue, and reason for the logs. Otherwise looks sane.

Yeah, in this case, right now, it looks logic to have that `move_issue()` function and save a bunch of lines... but that's only because the 2 current rules (A & B) are "move rules". For example, if you look to the continuous queues manager, that has way more rules added (hold, unhold, some with public comment and labels, others without...) then the actions become quite more heterogeneous.

Surely I'd leave that for now, so we can act on every rule as we want.

Ciao :-)

stronk7:
Ok, I've added a `run_C` that looks for issues awaiting integration as candidates too long (configurable, defaults to 14 days) and raises their integration priority to 1, so they will be picked before other issues when keeping the current queue fed.

Going to do some real testing now...

stronk7:
Ok, this is a dry-run with issues currently in the Tracker and config settings (currentmin = 15, movemax = 2, waitingdays = 2).

```
$ tracker_automations/normal_manage_queues/normal_manage_queues.sh 
Dry-run enabled, no changes will be performed to the tracker

Processing MDL-71659
Dry-run:  2021-05-20_23-43-02 MDL-71659 moved to current: important

Processing MDL-71659
Dry-run:  2021-05-20_23-43-02 MDL-71659 moved to current: threshold
Processing MDL-71644
Dry-run:  2021-05-20_23-43-02 MDL-71644 moved to current: threshold

Processing MDL-71628
Dry-run:  2021-05-20_23-43-02 MDL-71628 raised integration priority
Processing MDL-71159
Dry-run:  2021-05-20_23-43-02 MDL-71159 raised integration priority
```

So 1 important (I added the mdlqa label to it) issue is moved. Then, because we are under the threshold, 2 issues are moved. And finally, there are 2 issues that have been waiting > 2 days and they get the integration priority raised.

(note, you'll see that the MDL-71659 issue appears twice above, that's because that was a dry run and the issue fulfills 2 criteria to be moved, both important and under a threshold).

Ok, so now I'm going to perform a real run against the tracker, only the 4 issues actually in the candidates queue will be affected: MDL-71659, MDL-71644, MDL-71628, MDL-71159

stronk7:
Ok, these are the logs of the real execution:

```
 2021-05-20_23-50-30 [MDL-71659](https://tracker.moodle.org/browse/MDL-71659) moved to current: important
 2021-05-20_23-50-30 [MDL-71644](https://tracker.moodle.org/browse/MDL-71644) moved to current: threshold
 2021-05-20_23-50-30 [MDL-71628](https://tracker.moodle.org/browse/MDL-71628) moved to current: threshold
 2021-05-20_23-50-30 [MDL-71159](https://tracker.moodle.org/browse/MDL-71159) raised integration priority
```
(added links for easier visit)

Everything looks ok, all them have correct, restricted to integrators, messages explaining the action. The first 3 have been moved to current integration and the 4th has got its integration priority raised ([verified](https://tracker.moodle.org/issues/?jql=filter%3D14000%20AND%20%22Integration%20priority%22%20%3D1)).

So the 3 rules seem to be doing their (real) job as expected.

Ciao :-)

andrewnicols:
This gets my +1 then. I think that raising the priority in `run_C()` is the best approach here.

stronk7:
I think we can go ahead with this now and put it under production, future tweaks / rules can be added while we use it.

sarjona:
Thanks for working on this, Eloy. The patch looks good and I really appreciate the way you've documented it. It makes easier to read and understand it!
I like the way this "normal queues manager" will work!! Amazing idea and implementation!!

stronk7:
Thank you!

stronk7:
Issue description updated because points 4 and 6 are now done too.

PR #238, opened 2021-09-06
Prevent scripts against some buggy JAVA en_AU locales

stronk7:
From CLDR29 to CLDR33, the month and day names for en_AU were broken, incorrectly
adding a dot (.) after the day and month names. And it was fixed in CLDR34 . So I imagine
that java versions >=9 (when they changed from own locales to CLDR) that are using those CLDR
versions… are doomed (that for sure includes Java10).

Alternatively, there is the following switch:

`-Djava.locale.providers=COMPAT, CLDR`

to make Java own locales to be used instead of CLDR ones.

But as far as en_AU is already fixed in actual versions, we are simply
removing those incorrect dots fom the date fetched with Jira CLI and
done.

Some links:

- https://bugs.openjdk.java.net/browse/JDK-8208487

### DIFF
--- a/tracker_automations/count_delayed_last_cycle/count_delayed_last_cycle.sh
+++ b/tracker_automations/count_delayed_last_cycle/count_delayed_last_cycle.sh
@@ -76,6 +76,7 @@ done
 
 # get the contents of the integration date
 integrationjira=$( cat "${tempfile}" )
+integrationjira="${integrationjira//.}" # Some buggy Java AU locales come with dots in month names. Remove them.
 
 # if the last integration at jira has changed... start a new cycle
 if [ "${lastintegrationjira}" != "${integrationjira}" ]; then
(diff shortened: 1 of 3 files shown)


stronk7:
Note this is affecting all CI counters (reopened, delayed, tested...) because it uses date matching to detect when a new integration cycle begins.

Right now, with dates flip-flopping between `2/Sep./21` (incorrect) and `2/Sep/21` (correct), last integration dates are not matching and that is causing a new "counters cycle" to be started, discarding old issues.

With the normalization in the patch, that will stop happening (dots gone).

Ciao :-)

stronk7:
self-merged, we need the dates flip-flopping to stop ASAP because it kills counters.

First run with the patch applied is: https://ci.moodle.org/view/Tracker/job/TR%20-%20Count%20reopened%20issues/31118/console . There shouldn't be dates with dots in the month names anymore, let's see it in 24h...

PR #239, opened 2021-09-13
Add support to specify the runner to use

stronk7:
Defaulting to STABLE, also TESTING can be used.

Only to the manual "list_of_mdls" jobs (normal and security).

Also, detected some stuff pending in the security one, about
the ability to pick the jobs to execute, so reconciling it
to be feature-paired with the normal one.


andrewnicols:
LGTM. Feel free to self-merge when tests finish.

PR #241, opened 2021-09-23
Lower priority blocked

stronk7:
This includes 2 commits:

- the first one about to implement in the "change to zero" script the ability to lower the priority for any issue being blocked by unresolved issues.
- the second, about prevent the normal queues manager to perform any change of priority on issues being "blockers" or "blocked", that's something that the set_integration_priority_to_[zero|one] scripts manage in exclusive.

The commit messages do include more details.

Potential TODO: together with changing the integration priority, also add a restricted (to integrators) message from CiBoT explaining why it has been decided to raise or lower the priority (the queues manager already does it, but the set_integration_priority_to_[zero|one] scripts do not. Not really sure if it's needed, so just commenting about it to gather opinions. Than can be done apart in any case, it's 3 lines change.

### DIFF
--- a/tracker_automations/normal_manage_queues/lib.sh
+++ b/tracker_automations/normal_manage_queues/lib.sh
@@ -119,6 +119,7 @@ function run_C() {
     ${basereq} --action getIssueList \
                --search "filter=14000
                      AND 'Integration priority' = 0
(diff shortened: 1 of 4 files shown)


stronk7:
Thanks!

PR #242, opened 2021-10-13
Add classic runs to ToBiC jobs

stronk7:
With 4.0, boost is getting a lot of changes and we need to ensure
that classic continues ok. This adds a classic run to all ToBiC
runs (automated or manually invoked by integrators).


stronk7:
Just to note that the failures above are because of Travis sucking badly and not having the new Let's Encrypt CA certificates in their environment images.

That can be fixed apart by manually apt-installing the new certificates and openssl versions but, in any case, it's apart from this issue, so it won't be fixed here.

I'd consider the changes ok (in fact they aren't covered by the existing tests).

Ciao :-)

PR #244, opened 2021-11-02
Update to use GHA

andrewnicols:
Ditch Travis in favour of GHA.

Note: This issue is based on the issue to convert to https for github.

There are still a couple of failures here, but some of these look legitimate, and I'm not sure how best to handle the MySQL connectino failure issues just yet.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,93 @@
...
+      LOCAL_CI_TESTS_DBUSER: moodle
+      LOCAL_CI_TESTS_DBPASS: ""
+
+    services:
+      mysql:
+        image: mysql:5.7
+        env:
+          MYSQL_DATABASE: example
+          MYSQL_USER: ${{ env.LOCAL_CI_TESTS_DBUSER }}
+          MYSQL_PASSWORD: ""
```
scara, line 23: Hi @andrewnicols,
should it be:
```yaml
          MYSQL_PASSWORD: ${{ env.LOCAL_CI_TESTS_DBPASS}}
```
to be future-proof against any change about the idea to use empty passwords?

HTH,
Matteo

scara:
Hi @andrewnicols,
should MySQL issues be related to not using `MYSQL_ALLOW_EMPTY_PASSWORD=yes`?
Not sure, since reading https://github.com/docker-library/mysql/blob/15fe7357b165ee8aaa3ce165386f910a53a75087/5.7/docker-entrypoint.sh doesn't show any significant difference but "style" of writing the ENVs for MySQL docker - one ENV VAR instead of another.

HTH,
Matteo

.github/workflows/ci.yml
```diff
@@ -0,0 +1,93 @@
...
+jobs:
+  CI:
+    env:
+      LOCAL_CI_TESTS_CACHEDIR: ${{ GITHUB.WORKSPACE }}/cachedir
+      LOCAL_CI_TESTS_GITDIR: "${{ GITHUB.WORKSPACE }}/moodle"
+      LOCAL_CI_TESTS_PHPCS_DIR: "${{ GITHUB.WORKSPACE }}/moodle-local_codechecker/moodle"
+      LOCAL_CI_TESTS_DBLIBRARY: native
+      LOCAL_CI_TESTS_DBTYPE: mysqli
+      LOCAL_CI_TESTS_DBHOST: 127.0.0.1
+      LOCAL_CI_TESTS_DBUSER: moodle
```
scara, line 14: Hi @andrewnicols,
I wonder if that should be `root` since https://docs.travis-ci.com/user/database-setup/#mysql tells that their `travis` user is equivalent to `root` in their way to setup the MySQL service i.e. no need to create an unprivileged `moodle` user that will have all the powers but limited to the `example` DB - it will fail in creating a new DB (`ci_installed_6678_0`) - which is not the actual MySQL connection error (`ERROR 2003 (HY000): Can't connect to MySQL server on '127.0.0.1' (111)`).

HTH,
Matteo

scara:
Hi @andrewnicols,
I think that https://github.blog/changelog/2020-02-21-github-actions-breaking-change-ubuntu-virtual-environments-will-no-longer-start-the-mysql-service-automatically/ could be related.

HTH,
Matteo

stronk7:
Closing this in favour of #263 that has just landed.

PR #245, opened 2021-11-15
New upgrade_external_backup added to CiBoT

stronk7:
New check looking for missing modifications in external/backup when there are new columns added to DB in upgrade.

It happens "often" (here and there) that new columns are added to database in any component and it isn't checked if those new columns require changes in 2 critical areas: backup and external functions.

So, with this new check... when it's detected (in the whole patch being analysed) that there are new columns being added @ any `install.xml` and `upgrade.php` file... and there isn't any change to backup (`/backup/moodle2`) or external (`externallib.php|/external/`) some warnings are added, with text:

```
- WARN: Database modifications (new tables or columns) detected in the patch without any change to some important areas.
- WARN: No changes detected to external functions, that may affect apps and other web service integrations, please verify!
- WARN: No changes detected to backup and restore, that may affect storage and transportability, please verify!
```

Note that they are only warnings because it perfectly could be that everything is ok and the new columns don't require any change to backup or to any external. But it's warned in order to remind to everybody that it must be checked.

The new check (`upgrade_external_backup_check.sh`) is completely covered with tests and also cibot (`remote_branch_checker`), that uses it, is.

Note that we cannot verify that CiBoT messages to the tracker are correct. It will be done as soon as this is lands. But it should work basically ok (only some small adjustment may be needed).

Ciao :-)

### DIFF
--- a/remote_branch_checker/checkstyle_converter.php
+++ b/remote_branch_checker/checkstyle_converter.php
@@ -37,7 +37,7 @@ list($options, $unrecognized) = cli_get_params(
 
 if ($unrecognized) {
     $unrecognized = implode("\n  ", $unrecognized);
-    cli_error("Unrecognised options:\n{$Unrecognised}\n Please use --help option.");
+    cli_error("Unrecognised options:\n{$unrecognised}\n Please use --help option.");
 
 }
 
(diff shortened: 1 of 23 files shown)


stronk7:
Self-merged, so we can give it some runs

PR #247, opened 2022-01-18
Remove the travis checker from CiBoT

stronk7:
With everything and everybody already moved to GHA, it
doesn't make much sense to keep the travis checker running
on every CiBoT execution.

Hence, removing it from CiBoT (remote_branch_checker) and
associated tests. But keeping the shell script and its own
tests, in case they are useful for anybody.

Also, added TODO about the possibility of adding a GHA
checker (a simple https request can tell us if actions are
enabled for a repo), but not sure if it's worth it as far
as 99% of people have them enabled.

### DIFF
--- a/remote_branch_checker/checkstyle_converter.php
+++ b/remote_branch_checker/checkstyle_converter.php
@@ -41,7 +41,7 @@ if ($unrecognized) {
 
 }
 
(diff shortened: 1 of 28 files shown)


sarjona (approved):
Thanks for working on this patch. It looks good so I'm happy to approve and merge it! :-)

PR #249, opened 2022-06-07
Verify phpunit xml

stronk7:
This light job does the following:

* Creates the phpunit.xml file (without requiring a whole site or database to be installed).
* Validates that all the "tests" directories in code base are covered by some suite in the phpunit.xml file. Errors if not (usually missing subsystem in phpunit.xml.dist file).
* Also reports (without failure) other "abnormal" situations like:
  * suite not having any _test.php file.
  * _test.php file not having any known unit test class (may require to inform about new parent class to the script).
  * _test.php file having multiple known unit tests classes. Warns by default, but can be configured to error too.

And that's all. Planning to add it to the post-checks @ integration.moodle.org, because it's quick and rarely happens (only when some new subsystem arrives). Like it has happened with 'course/format/tests" in 4.0. <= Will create issue as soon as this is up and running @ CIs.

Ciao :-)

### DIFF
--- a/phplib/clilib.php
+++ b/phplib/clilib.php
@@ -234,6 +234,7 @@ function load_core_component_from_moodle($moodledirroot) {
     $CFG->dirroot = $moodledirroot;
     $CFG->libdir = $CFG->dirroot . '/lib';
     $CFG->admin = 'admin';
+    $CFG->filepermissions = '0666';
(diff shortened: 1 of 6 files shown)


stronk7:
Tests passed, self-merging, in order to create the job ASAP.

PR #250, opened 2022-06-28
All jobs moving issues to current integration must set clr = no

stronk7:
All the issues arriving to current integration must have
the "Component Lead Review" field set to no, no matter if
the move is done by any of the queue managers or the bulk-move
that is executed on freeze day.

If they are kept empty / not set, that affects pickers because
one of the sorting criteria is that field.

Technically, all the places that are setting:

customfield_10211:Yes (Currently in integration = Yes)

Must, unconditionally, set:

customfield_15810:No (Component Lead Review = No)

That's exactly what this change does.


andrewnicols (approved):

stronk7:
Ta!

PR #253, opened 2022-08-09
MDLSITE-6709 Avoid running the backup & external check for plugin runs

stronk7:
As far as we don't check the plugins incrementally (commit by commit)
like we do for normal MDL issues by CiBoT or PRs by moodle-plugin-ci,
this check is not useful for checks coming from the plugins database
(because they are just standard moodle branches + 1 unique commit
with all the plugin source code).

So we are disabling this check to be run when the source branches
are PLUGIN-xxx ones (we do the same for the commit check already).

Covered with tests.

PS: Also, delete old TODO comment that should have been deleted
few weeks ago, when another PR was merged.

Note this fixes/closes https://tracker.moodle.org/browse/MDLSITE-6709


stronk7:
I'm self-merging this, it's trivial and tested, so it should be safe enough.

PR #255, opened 2022-09-13
Move from days to hours for timing better resolution

stronk7:
We need to use hours instead of days, because the "AFTER -5d" is not accurate enough. In practice it accounts for 4.5 days, or something like that, hence the actions are happening some hours before they should. So, using hours (days * 24) to get more accurate results.

See some of the already processed issues like MDL-66273, MDL-74993 or MDL-75278... it's easy to see there how the reminder notification is happening after 4.5days (aprox), but before the 5 days configured.

Changing to hours will be way more accurate (max error will be just half an hour, not half a day). Note we already use this in other tracker automations, because of the accurateness.

Ciao :-)

### DIFF
--- a/tests/libs/shared_setup.bash
+++ b/tests/libs/shared_setup.bash
@@ -128,12 +128,12 @@ restore_workspace() {
 # Clever idea, borrowed from https://github.com/dgholz/detect_virtualenv/blob/master/t/begin_and_end.bash
 # Are we in the first test of a file?
 function first_test() {
-  [ "$BATS_TEST_NUMBER" = 1 ]
+  [ "$BATS_TEST_NUMBER" -eq 1 ]
 }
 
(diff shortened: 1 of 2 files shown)


stronk7:
Uhm... the failure is happening since some weeks ago... investigating it... but it's 100% unrelated to the changes in this PR... looking...

stronk7:
Ok, I think I've found which the changed (internal to bats-core) are... and found a valid workaround to keep those tests (detecting the last test in a file) working.

Have added a commit to this branch incorporating the changes, let's see if that makes Travis happy.

Ciao :-)

stronk7:
Ok, tests passed now. I'm self-merging this, because we have some ongoing issues that will be a good testing table for the changes performed here. Now notifications should happen closer to the 5 & 7 days.

Ciao :-)

PR #256, opened 2022-09-15
Add renamed classes following MDL-75716

andrewnicols:
In MDL-75716, unit test class names were updated to match file names. As a result the parent class for a couple of tests was changed and must be added to the list of known parent classes.

The old location is not removed as they may still be legitimately used in older versions of Moodle.

### DIFF
--- a/verify_phpunit_xml/verify_phpunit_xml.sh
+++ b/verify_phpunit_xml/verify_phpunit_xml.sh
@@ -65,6 +65,7 @@ unittestclasses="
     mod_assign\\\\externallib_advanced_testcase
     mod_lti_testcase
     mod_quiz_attempt_walkthrough_from_csv_testcase
+    mod_quiz\\\\attempt_walkthrough_from_csv_test
     provider_testcase
     qbehaviour_walkthrough_test_base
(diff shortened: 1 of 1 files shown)


verify_phpunit_xml/verify_phpunit_xml.sh
```diff
@@ -65,6 +65,7 @@ unittestclasses="
     mod_assign\\\\externallib_advanced_testcase
     mod_lti_testcase
     mod_quiz_attempt_walkthrough_from_csv_testcase
```
junpataleta, line 67: Should we remove this one as well?

junpataleta (commented):
Thanks, Andrew. Looks good. Just wondering if we should remove the old class name `mod_quiz_attempt_walkthrough_from_csv_testcase` as well.

verify_phpunit_xml/verify_phpunit_xml.sh
andrewnicols: No. See commit messages 🤣
We still run this on older branches and those outside our control.

junpataleta:
Thanks! Merged!

PR #257, opened 2022-09-28
Fix the reopen timing and notification

stronk7:
Because the reminder, sent after 5 days (for example) makes a fake transition (from WfF to WfF) we need to account separately for issues without changes:
    - Since sent to WfT  (7 days, for example) for issues that weren't reminded (notification = 2)
    - Since remind happened (2 days, for example) for issues that were reminded (notification = 3)


stronk7:
See for example [MDL-75032](https://tracker.moodle.org/browse/MDL-75032), it was reopened too late (7 days after the reminder, when it should have been reopened 2 days after the reminder).

And, once this is landed, see for example [MDL-75497](https://tracker.moodle.org/browse/MDL-75497), it should be immediately reopened, because it's already 2 days after the reminder.

Ciao :-)

andrewnicols (approved):

PR #258, opened 2023-01-11
Move away from MaxCDN to CloudFlare CDN

stronk7:
MaxCDN has stopped serving the twemoji images, see:

https://github.com/twitter/twemoji/issues/580

So we are moving to cdnjs.cloudflare.com here so
all those tiny images used by CiBoT (some day we'll converted the majority of them to just Unicode emojis) continue working and making devs to smile a little bit.

This is a test to check that all them are working ok:

pompoms balloon popper beer cookie cake star smile

fire poo monkey bug fear construction sos skull

### DIFF
--- a/tracker_automations/bulk_precheck_issues/util.sh
+++ b/tracker_automations/bulk_precheck_issues/util.sh
@@ -32,19 +32,19 @@ function get_happy_image() {
     # marys pompoms (from https://tracker.moodle.org/browse/MDLSITE-3642)
     images[0]='https://tracker.moodle.org/secure/attachment/46622/AnimatedAmy.gif'
     # balloon🎈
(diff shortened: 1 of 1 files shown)


andrewnicols (approved):

andrewnicols:
Merged ignoring Travis failures which are clearly unrelated.

stronk7:
Thanks! Will look to the failure, it's curious why the DB comparison job is reporting more differences (7) than the expected ones (6).

stronk7:
Ok,

in the test we are comparing an installed 4.0dev with an upgraded 311_STABLE. And the problem is that, in [MDL-75766](https://tracker.moodle.org/browse/MDL-75766) we backported one index that is not in the 4.0dev version.

#259 should fix that problem, cross fingers.

Ciao :-)

PR #262, opened 2023-03-16
Change to PGSQL temporarily until we fix the recent issues with sqlsrv

lameze:
We have been getting some failed builds with database errors on sqlsrv runs, see https://github.com/moodlehq/moodle-php-apache/pull/164#issuecomment-1472883622

This commit changes the ToBiC and awaiting_for_integration runs to use pgsql database temporarily.


lameze:
Andrew has fixed

PR #263, opened 2023-03-17
Moving from Travis to GHA

stronk7:
Not much to say, one less repository using Travis.

It comes with 2 extra commits (both have details in the commit messages):

- One about ensuring that we avoid removing the correct directories only (detected because there was a conflict with GHA own paths).
- Another about to enforce commit hashes to be 12 chars long. We are right now on the edge of switching from current 10 chars to 11 chars (that's calculated automatically by git given the number of objets in the repository), so let's force 12 to be used (that should last some good years, before 13 are needed).

This is the last repo remaining to be moved to GHA, only core and moodle-plugin-ci once this is accepted. Yay!


stronk7:
Self-merging, so we get GHA enabled and Travis off.

PR #264, opened 2023-03-27
Fail the global CiBoT status if unsupported GH git:// URLs are used

stronk7:
No matter of that, the checks still will happen, but it's required to the developer to amend the "Repository" field in the issue.


stronk7:
Self merging, this has been already tested in place ([MDL-73226](https://tracker.moodle.org/browse/MDL-73226)) and it's working ok.

PR #266, opened 2023-05-22
Ensure that the grunt job also runs non-default tasks

stronk7:
This includes:
- ignorefiles (that was already being checked, no matter it's called by other tasks).
- jsconfig
- jsdoc
- upgradablelibs

Covered with a real test using Moodle v4.2.0, where jsdoc was failing and nobody noticed.


stronk7:
Thanks, Andrew! Merging...

PR #267, opened 2023-06-26
Bump vnu-jar to latest

andrewnicols:
We have had reports of valid HTML being rejected because it was not valid at the time of the release of the version of vnu-jar we were using.

The previous version was from 2018.

Bumping to latest - 2023.4.11

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,13 +8,13 @@
       "name": "moodle-local_ci",
       "version": "0.0.1",
       "dependencies": {
-        "vnu-jar": "18.11.5"
+        "vnu-jar": "23.4.11"
       }
     },
     "node_modules/vnu-jar": {
-      "version": "18.11.5",
-      "resolved": "https://registry.npmjs.org/vnu-jar/-/vnu-jar-18.11.5.tgz",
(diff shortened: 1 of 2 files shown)


junpataleta (approved):
Looks spot on. Thanks!

ewallah:
Just for the record, but since this change, themes based on boost now receive warnings in mustache:

`
WARNING: HTML Validation info, line 5: Trailing slash on void elements has no effect and interacts badly with unquoted attribute values. (<link rel="shortcut icon" href="favicon.ico" />)
`

The [same warning](https://github.com/moodlehq/moodle-local_ci/actions/runs/5373102304/job/14540963283?pr=267#step:11:79) that fails test 1-* in Github Actions.

So the theme/boost/templates/head.mustache file needs a fix...

andrewnicols:
Have you created an issue in the Moodle tracker?

This warning is correct, so it needs to be fixed in Moodle.

ewallah:
Nope, not yet done.  Because I think this problem also will pop up in other parts of Moodle (yui modules - fontawesome - tinymce -....), I was searching for a more general approach. Any ideas?

As a workaround I added this to my workflow file:

```
      - name: fix mustache
        run: sed -i 's/ \/>/ >/g' head.mustache
        working-directory: moodle/theme/boost/templates

```

ewallah:
Issue [MDL-78809](https://tracker.moodle.org/browse/MDL-78809) created.

PR #272, opened 2023-07-11
Ensure the str helper is url/href compliant

### LINKED ISSUE #270: Validating mustache templates that use string helpers for URLs causes warnings
We use language strings in a certain way to improve accessibility:

```mustache
<a href='{{#str}} localised_url, local_plugin {{#str}}'>some external page</a>
```

The localised URL is usually something like `example.org/?lang=de` or `example.com/en`.

Unfortunately, [`moodle-local_ci/mustache_lint/mustache_lint.php`](https://github.com/moodlehq/moodle-local_ci/blob/4bc8230ff53b341cf3fa14ac5fd2ae37d274ee57/mustache_lint/mustache_lint.php#L90) wraps the helper’s parameters in brackets (`[[ localised_url, local_plugin ]]`) which causes warnings when validating the HTML, because brackets are not allowed in a URL.

Could we maybe just `urlencode()` the string instead of wrapping it in brackets?  
What was the reasoning behind using brackets in the first place?  
Maybe there is an even better way of solving this?

stronk7:
So we are:
- replacing the square brackets by parenthesis (allowed in URLs).
- urlenconding the str contents so they are also allowed links.

Backed by tests.

Fixes #270

### DIFF
--- a/mustache_lint/mustache_lint.php
+++ b/mustache_lint/mustache_lint.php
@@ -87,7 +87,7 @@ $jshelper = new js_helper($CFG->dirroot);
 $mustache = new Mustache_Engine([
     'pragmas' => [Mustache_Engine::PRAGMA_BLOCKS],
     'helpers' => [ // Emulate some helpers for html validation purposes.
-        'str' => function($text) { return "[[$text]]"; },
+        'str' => function($text) { return '((' . urlencode($text) . '))'; },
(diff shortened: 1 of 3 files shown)


stronk7:
I'm self merging this towards an imminent release to be used by others (moodle-plugin-ci, CiBoT and friends...)

PR #274, opened 2023-07-17
Treat Mustache "info" message as info, not problem

james-cnz:
The Mustache lint seems to treat all HTML validation messages as problems, even if they are "info".

I have run into this recently, while checking my course format plugin.  I guess there's been a recent change to the HTML validator that's caused it?  A Mustache file failed validation because some included Moodle Mustache files closed HTML single tags with /> instead of just >.  I got the "info" message "Trailing slash on void elements has no effect and interacts badly with unquoted attribute values."  As far as I can tell, this message is only ever given when there is no actual problem, and if there is an actual problem with a trailing /, I think another, non-info message will likely be given.

I don't think it's necessary to treat "info" messages as problems, and I suspect it will now cause most course format plugins to fail validation in a way that's beyond the author's control.  (Not sure about other plugin types.)

I _think_ the suggested change should fix the issue, but I haven't actually figured out how to run it locally to check, so I don't really know.

I assume this plugin is used for GitHub actions, and the Moodle code prechecks?  If so, and this change is accepted, could you tell me how long it's likely to be before those use the update?

### DIFF
--- a/mustache_lint/mustache_lint.php
+++ b/mustache_lint/mustache_lint.php
@@ -237,7 +237,11 @@ function check_html_validation($content) {
         //TODO: Sure would be nice if we could 'guess' a more accurate line number here..
         $context = str_replace("\n", '', $problem->extract);
         $message = "HTML Validation {$problem->type}, line {$problem->lastLine}: {$problem->message} ({$context})";
-        print_problem('WARNING', $message);
+        if ($problem->type == 'info') {
+            print_message('INFO', $message);
+        } else {
+            print_problem('WARNING', $message);
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


james-cnz:
Sorry, I thought I'd run the tests in my fork, but I guess I accidentally ran them against the master branch, rather than my change.  Should be right now.

stronk7:
Hi @james-cnz ,

thanks for the PR and your patience! I started looking to this some days ago, because I had to test other VNU stuff, but had to stop and haven't been able to come back to this till now.

Looking...

stronk7:
Hi again,

I've amended slightly your test (so it continues looking also for errors in the same file), have forced-pushed to your repository.

Once the GHA tests end here (hopefully passing), I'll merge this.

Ciao :-)

PR #275, opened 2023-07-26
Move issues to the current queue by Rank

junpataleta:
In order for the current integration queue to reflect the ordering of issues in the integration backlog, the queues manager will now be moving issues by `Rank` instead of by `Integration priority` and then `Rank`.

It will be up to the product managers to rank all issues regardless of the issues' Integration priority. But it will still be highly encouraged to rank issues with `Integration priority=1` higher whenever possible.


ilyatregubov:
Thanks Jun! Merged

PR #276, opened 2023-08-04
Various changes towards CI passing with php82

stronk7:
Right now we are running local_ci various scripts with php74 and php80, and it can be continue being used that way for years. But, as part of the [PHP 8.2 Epic](https://tracker.moodle.org/browse/MDL-76405), where a bunch of our tools (codechecker, moodle-docker...) are modified to confirm that they work perfectly with the new PHP version, I decided to give to local_ci a run.

It has not been easy, but now it's passing and the few places in the php scripts needing fixes have been detected.

- Fixed a few php82-isms.
- Changed some expectations because php lint messages are different.
- Use custom fixture to test that the compare_databases job detects problems.
- Restrict tests to use only 402 and master.
- Added some TODOs about to use more stable branches in the future once we get more supporting php82.

So, we are green now with php82, that's all. Achieved.

Ciao :-)


stronk7:
Ty!

PR #277, opened 2023-08-09
Don't try any rebase --onto when both repos have matching HEADs

stronk7:
If both integration.git and security.git, for a given branch, have already the same HEADs, then we don't need to rebase anything, because they are already the same.

The only option to perform is to update the -lastbased branch to point to the same HEAD and forget.


stronk7:
Tested locally against 400_STABLE and it has fixed the problem that happened immediately after running the integration2security process and having some security fixes having multiple, overlapping commits (that cause the rebase --onto to fail with conflicts).

junpataleta (approved):
Nice one. Thanks, Eloy!

PR #278, opened 2023-08-29
Enable back (partially) the move from candidates to current

stronk7:
On continuous we want these "important" issues to be automatically moved to current integration, irrespectively of the "threshold" moves:
- The issue has a must-fix version. [OR]
- The issue has the mdlqa label.

Of course, only issues in the candidates queue, not held and not blocked by others.

Note that we have removed some of the old "important" criteria, namely:
- The issue priority is critical or higher.
- The issue is flagged as security issue.
- The issue belongs to some of these components:
  - Privacy
  - Automated functional tests (behat)
  - Unit tests

See [MDLSITE-7296](https://tracker.moodle.org/browse/MDLSITE-7296) for more info (comments around 29 August 2023).

Tested with [MDL-70371](https://tracker.moodle.org/browse/MDL-70371), seems to be working ok (that issue was a must-fix).


andrewnicols (approved):

stronk7:
Thanks!

PR #279, opened 2023-08-30
Add support to run just Chrome jobs

stronk7:
By using the new "behat-chrome" jobtype, all the
configurable criteria now will launch chrome DEV jobs.

Note that Firefox continues being the default (more reliable) browser used by all criteria.


stronk7:
Thanks!

PR #282, opened 2023-10-09
Make all the "holding" actions to apply both for IR & CLR

stronk7:
Now, all the actions related to holding issues:
- After freeze, any new feature or improvement.
- Last week, any issue type.
- While on-sync, any new feature and improvement.

Are managed both for issues awaiting for IR and awaiting for CLR, so the same outcome (hold issues) happen
exactly the same for both.

Note that we'll have to modify the manual release process slightly to remember to unhold issues from both queues, that will be documented @ https://moodledev.io/general/development/process/release once this is merged and working.

### DIFF
--- a/tests/1-compare_databases.bats
+++ b/tests/1-compare_databases.bats
@@ -108,18 +108,18 @@ teardown () {
 }
 
 @test "compare_databases/compare_databases.sh: problems are detected" {
-    # Locally, patch master (4.3dev), as of 20230803, so we introduce some differences.
-    create_git_branch compare_databases  a1d5d1b2f758a3178bff4c42883a477f4b1ea50c
+    # Locally, patch v4.3.0, so we introduce some differences. Then compare with upgraded v4.2.2.
+    create_git_branch v4_3_0_wrong v4.3.0
     git_apply_fixture compare_databases_wrong.patch
 
-    export gitbranchinstalled=compare_databases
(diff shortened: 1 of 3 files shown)


stronk7:
The [failure above is really curious](https://github.com/moodlehq/moodle-local_ci/actions/runs/6454855509/job/17521210438?pr=282#step:11:44), not sure how it can detect a "downgrade" when the fixture that we are using only changes a few database column specs to verify that differences are detected...

Going to relaunch the job and investigate... but for sure **it's unrelated with this PR** (that only touches the continuous queues manager).

Ciao :-)

stronk7:
Ok, I've added one more commit that fixes a bogus test (that started happening as soon as we made a new release in the 402_STABLE branch).

The test has been modified to always run between 2 already release versions (v4.2.2 and v4.3.0), so it will not affected by future releases any more.

Crossing fingers to get GHA green now...

Ciao :-)

sarjona (approved):
Hi Eloy!
Thanks for creating a patch for fixing these issues and improving our processes! It looks good :-)

stronk7:
Mer~~ged~~CI !!! :-P

stronk7:
Side note, we don't need to document in the release process the un-holding of the CLR issues because it was already documented, yippee! 

Reference: https://github.com/moodle/devdocs/pull/476

Ciao :-)

PR #283, opened 2023-10-10
Mustache helper now trim JS contents before passing them to ESLint

stronk7:
Now the JS helper is trimming JS contents
    
This solves some false positives when the mustache checker is validating (using eslint) all the `{{#js}}...{{/js}}` sections embedded within templates.

Added a new test covering a real case in v4.3.0 of some partial with white space around.
    
Also we have added a bit of extra information to output so we can assert better in tests if the ESLint check happened or no. Various tests are now looking for that new output.


stronk7:
Once this is accepted, I'll be making a new `moodle-local_ci` "release" (tag), and then will proceed to also release `moodle-plugin-ci` because this is impacting to plugins using some, used often, core partials.

stronk7:
(self-merging...)

PR #285, opened 2023-11-15
GHA is deprecating Node 16 actions, let's bump them

stronk7:
Special note about the Bats action, we are moving
to a new one because the actual is not maintained.

amdilabs/setup-bats@v1 ==> brokenpip3/setup-bats-libs@1.5.1

Let's see how it goes...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,12 +15,12 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Checking out moodle-local_ci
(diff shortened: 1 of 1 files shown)


stronk7:
For reference, needed a few adjustments to the `brokenpip3/setup-bats-libs@1.5.2` GHA to make it work without any warning (problem with GHA caching and default locations in the action).

Link: https://github.com/brokenpip3/setup-bats-libs/issues/18

stronk7:
Self-merging, so I can continue moving some tests...

PR #286, opened 2023-11-20
Fixes the generation of displayname when integrateto is set to main

snake:
This is currently setting an empty string when 'integrateto' is set to 'main', causing tools like mdk precheck to fail. E.g. 
https://integration.moodle.org/view/prechecker/job/Precheck%20remote%20branch/148255/console

That job uses MDL-79114-master as the branch name, but 'integrateto' is set to 'main' (see the job params), which results in the job failure. Running a job manually works fine, as long as integrateto is set to master.


stronk7:
Out of curiosity… that implies that MDK is dependent of the job display name?

stronk7:
Merged, thanks @snake ! Hope it fixes the MDK problems... I'll be reviewing the whole CiBoT script later this week.

PR #290, opened 2024-01-17
Prevent plugin runs to use the moodle.Commenting.TodoComment sniff

### LINKED ISSUE #289: Ensure that we only use the TodoComment Sniff for core runs, not for plugin ones
With https://github.com/moodlehq/moodle-cs/pull/91 we are introducing a new Sniff that looks for missing MDL-xxxx issue codes in all the TODO and @todo comments (as stated by the coding style).

This Sniff is only for core, hence we have to prevent it to be executed for plugins within CiBoT. By disabling it or by setting `commentRequiredRegex` to an empty value.

Ciao :-)

stronk7:
That sniff covers the core coding style and verifies that all the TODOs (inline or phpdoc block), have a MDL issue associated.

But that's not a requirements for plugins, hence disabling that Sniff for them.

Fixes #289


stronk7:
Slef-merging this, so we have local_ci (and CiBoT) ready for the changes coming in moodle-cs.

PR #294, opened 2024-03-12
Validate release date on continuous queues manager

### LINKED ISSUE #293: Validate release date before performing actions on the `TR - Manage queues on continuous` job
If the integrator enables the `TR - Manage queues on continuous` job but forgets to update the `Release date` field to the correct value (a.k.a. me), the Continuous queues manager will not work properly and can even post the incorrect message when holding issues (e.g. posting a hold message on held issues for the on-sync period).

We must ensure the release date configured on the job is valid before processing. Otherwise, the job should fail with an error about the invalid release date configured on the job. A probable validation logic is to simply ensure that the current date is before the end of the on-sync period for the configured release date.

junpataleta:
This fixes #293

### DIFF
--- a/tests/3-continuous_manage_queues_validation.bats
+++ b/tests/3-continuous_manage_queues_validation.bats
@@ -1,0 +1,88 @@
+#!/usr/bin/env bats
+
+load libs/shared_setup
+
+setup () {
+    export WORKSPACE=$gitdir
+
+    # Some dummy variables just for the sake of the letting the script proceed through the validation.
+    export jiraclicmd=jiracli
+    export jiraserver=https://tracker.moodle.org
+    export jirauser=jirauser
+    export jirapass=jirapass
+}
+
+@test "tracker_automations/continuous_manage_queues/continuous_manage_queues.sh: Current date < release date" {
+    releasedate=$(date -d "+10day" +%Y-%m-%d)
+    lastweekdate=$(date -d "${releasedate} -7day" +%Y-%m-%d)
+
+    source $PWD/tracker_automations/continuous_manage_queues/lib.sh
+
+    run run_param_validation $releasedate $lastweekdate
+
+    # Assert result.
+    assert_success
+    assert_output --partial "Parameters validated"
+}
+
+@test "tracker_automations/continuous_manage_queues/continuous_manage_queues.sh: Current date = on-sync date" {
+    releasedate=$(date -d "-28day" +%Y-%m-%d)
+    lastweekdate=$(date -d "${releasedate} -7day" +%Y-%m-%d)
+
+    source $PWD/tracker_automations/continuous_manage_queues/lib.sh
+
+    run run_param_validation $releasedate $lastweekdate
+
+    # Assert result.
+    assert_success
+    assert_output --partial "Parameters validated"
+}
+
+@test "tracker_automations/continuous_manage_queues/continuous_manage_queues.sh: Current date > on-sync date" {
+    releasedate=$(date -d "-29day" +%Y-%m-%d)
+    lastweekdate=$(date -d "${releasedate} -7day" +%Y-%m-%d)
+
+    source $PWD/tracker_automations/continuous_manage_queues/lib.sh
+
+    run run_param_validation $releasedate $lastweekdate
+
+    # Assert result.
+    assert_failure
+    assert_output --partial "ERROR: The current date is already past the on-sync period. Please make sure the Release date (${releasedate}) is configured correctly"
+}
+
+@test "tracker_automations/continuous_manage_queues/continuous_manage_queues.sh: Invalid release date format" {
+    releasedate=$(date +%m-%d-%Y)
+    lastweekdate=$(date +%Y-%m-%d)
+    source $PWD/tracker_automations/continuous_manage_queues/lib.sh
+
+    run run_param_validation $releasedate $lastweekdate
+
+    # Assert result.
+    assert_failure
+    assert_output --partial "ERROR: \$releasedate. Incorrect YYYY-MM-DD format detected: ${releasedate}"
+}
+
+@test "tracker_automations/continuous_manage_queues/continuous_manage_queues.sh: Invalid last week date format" {
+    releasedate=$(date -d "+7day" +%Y-%m-%d)
+    lastweekdate=$(date +%m-%d-%Y)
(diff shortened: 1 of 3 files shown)


stronk7:
Thanks @junpataleta , and covered with tests, yay!

It looks perfect, will merge as soon as all checks are done, hopefully green.

Ciao :-)

stronk7:
As commented via chat, I think that we have to remove the 2 "dry-run" tests. We cannot test that because they require the tracker to be available (for querying). And it's not, so they are failing, and bats is warning us about that (because they don't have any `assert_success` or `assert_failure`.

So, the 2 solutions are:
1. or we remove the dry-run tests.
2. or we add `assert_failure` to them, because they are going to fail yes-or-yes (I imagine).

Personally, I'd remove them, because we are not going to be able to test the script without a real tracker, so no query or update can be tried ever.

Ciao :-)

junpataleta:
Thanks, @stronk7. I went with the approach of moving the parameter validation to its own function so it can be tested separately. I tidied up the patch. Please have another look.

tracker_automations/continuous_manage_queues/continuous_manage_queues.sh
```diff
@@ -72,21 +72,11 @@ movemax=${movemax:-3}
...
-# Verify that $lastweekdate has a correct YYYY-MM-DD format
-if [[ ! ${lastweekdate} =~ ^[0-9]{4}-[0-9]{2}-[0-9]{2}$ ]]; then
-    echo "ERROR: \$lastweekdate. Incorrect YYYY-MM-DD format detected: ${lastweekdate}"
-    exit 1
-fi
-
 # Today
 nowdate=$(date +%Y%m%d)
 
+run_param_validation $releasedate $lastweekdate
```
stronk7, line 78: As far as you're relying on this script being executed with `-e` to exit on functions returning non-zero... it's a risk if, someday, that's changed for some reason (say debugging or whatever). So I would recommend to do some of these, to ensure that we don't forget about the script requiring `-e`.
- Or you add a comment in the line saying that this script won't work if changed.
- Or you change the `return 1` statements by direct `exit 1` calls.
- Or you capture the results of the function and add a "guard" condition to, in case it returns non-zero, it exits.

As said, the current code is correct, but we need to ensure that the behavior is the same as much as possible, no matter of the exit-on-error setting.

tests/3-continuous_manage_queues_validation.bats
```diff
@@ -0,0 +1,68 @@
...
+    export jirapass=jirapass
+}
+
+@test "tracker_automations/continuous_manage_queues/continuous_manage_queues.sh: Current date < release date" {
+    export releasedate=$(date -d "+10day" +%Y-%m-%d)
+    source $PWD/tracker_automations/continuous_manage_queues/lib.sh
+
+    run run_param_validation $releasedate
+
+    # Assert result.
```
stronk7, line 21: I would add `assert_success` (maybe because I always add one of success/failure assertions). That way, if some error happens later, after the message has been printed, it will be detected.

stronk7 (changes requested):
We are 99% there, great refactor and tests! I've added a couple of comments in the patch...

tracker_automations/continuous_manage_queues/lib.sh
```diff
@@ -389,3 +389,30 @@ function run_C() {
         echo "$BUILD_NUMBER $BUILD_TIMESTAMP ${issue} moved out from current: held" >> "${logfile}"
     done
 }
+
+function run_param_validation() {
+    nowdate=$(date +%Y%m%d)
+    releasedate=$1
+    lastweekdate=${2:-$(date -d "${releasedate} -7day" +%Y-%m-%d)}
```
stronk7, line 396: Ah, sorry, forgot to comment as part of the review. What if we do the 2 params mandatory? I mean, if any of them is empty... then return 1 too. That way, we can remove the code (dupe) recalculating the `lastweekdate`. Whoever is calling to the function will have to provide both dates.

For your consideration, I'm not strong on this (just hate dupe code, easy to forget if changes are performed, heh, maybe you need another function to calculate that date, that would be another alternative).

Ciao :-)

tracker_automations/continuous_manage_queues/continuous_manage_queues.sh
junpataleta: Thanks, Eloy! Good catch! I used `exit 1` for the failed validation cases.

tests/3-continuous_manage_queues_validation.bats
junpataleta: Ooops, yes. I added `assert_success`.

tracker_automations/continuous_manage_queues/lib.sh
junpataleta: Yep, good one. This is done.

junpataleta:
Thanks for your patience to a bash/bats naab like me, @stronk7 😅. I have made the changes you mentioned. 

I also added validation to ensure that `lastweekdate` is not after the `releasedate`.

I hope this is good to go.

stronk7 (approved):

stronk7:
Thanks, @junpataleta both for the PR and your patience! Hope it has been a good experience, heh!

About to merge this now, yay!

PS: We still have to discuss if we want, before enabling this script, to have a job to do both:
- Move IR issues from candidates to current (right now done by the `TR- Move awaiting...` and documented).
- Avoid CLR issues, that don't have the candidates/current split to be held (this time we have done it adding the unheld standard comment). There isn't a job for this, neither it's documented.

Notes:
- It can be a new job only for the CLR issues, or we can join both actions together.
- If the "adding the unheld standard comment" is not elegant solution, we can always create some field or add a more specific comment (sort of "This issue made the freeze while sitting in the CLR queues" or so). Then we can edit the filters to also avoid issues with that other comment.

Just using this PR to have my thoughts shared, heh.

Ciao :-)

junpataleta:
Yay! Thanks, @stronk7!

I think that the the continuous queues manager should be handling this task of holding or keeping waiting for CLR issues.

About preventing CLR issues from being held, would it be possible to get the date when the waiting for CLR issue has been sent to integration? For example, fetch the date when the issue's status changed from `Peer review in progress` to `Waiting for integration`.  

Then if it's determined that the date/time the issue was sent to integration falls before the code freeze, it will not be held. Otherwise, it will be held by the continuous queues manager. 

What do you think?

PR #297, opened 2024-05-16
Add a new "overview" section to the smurf report 

stronk7:
It will include all the warnings and errors that have been
detected along the execution of the checks.

Before this, we are handling those errors and warnings apart,
by the caller (bulk_precheck_issues) scripts. And that leads
to some extra complications to re-evaluate the final results
of the whole thing.

Now, those problems will be part of the smurf report and will
be handled like any other part of the report (phpcs, grunt, ...)
perfectly accounting for the final results and so on.


stronk7:
I’m self-merging this now, to give to it some runs at CI…

PR #301, opened 2024-05-29
Add autoloader to verify_phpunit_xml

andrewnicols:
Tested locally.
I'm not sure how I can add a test for this in bats because it's only broken by a change in core which requires an autoloaded class in lib/outputcomponents.php. We could switch from 3.11 to main for testing?

### DIFF
--- a/tests/verify_phpunit_xml.bats
+++ b/tests/verify_phpunit_xml.bats
@@ -13,7 +13,7 @@ teardown () {
     last_test || store_workspace
 }
 
-@test "verify_phpunit_xml: normal run without errors" {
+@test "verify_phpunit_xml: normal run without errors (old branch)" {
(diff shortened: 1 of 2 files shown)


stronk7:
Thanks Andrew, this is exactly what I was thinking about. Will look to add/modify the test soon. I think it’s easy to add one in this case pointing to main (no matter that the problem is still only @ integration.git).

stronk7:
Hi,

I've added an extra commit to your branch that ensures that the tool continues working with `main`. Note that it's useless right now (because moodle.git doesn't have the new, breaking, code yet), but it will be useful after rolling.

This has +1, feel free to merge it if you agree.

Ciao :-)

stronk7 (approved):

stronk7:
Merging this on @andrewnicols 's behalf :-) (and because I need this landed to provide another, hopefully interesting, PR.

PR #306, opened 2024-06-27
Stop autoloading lib/outputcomponents.php if the classes already exist

andrewnicols:
MDL-82183 moved all of the output classes to support autoloading and added exceptions to prevent manual inclusion of these files.


junpataleta (approved):
Thanks, Andrew! Spot-on patch!

PR #307, opened 2024-07-21
Report list of files from merge conflicts

ziegenberg:
When we experience merge conflicts in the CiBot Precheck it would be nice to have more information than just: "does not apply clean", e.g. https://integration.moodle.org/job/Precheck%20remote%20branch/160991/artifact/work/errors.txt/*view*/

This adds the output from `git diff --name-only --diff-filter=U`, which provides the list of files not merged because of a conflict.

### DIFF
--- a/remote_branch_checker/remote_branch_checker.sh
+++ b/remote_branch_checker/remote_branch_checker.sh
@@ -255,7 +255,7 @@ if [[ ${exitstatus} -ne 0 ]]; then
         echo "Error: Merge conflict(s) in file(s):" | tee -a ${errorfile}
         echo "${mergeconflicts}" | sed 's/^/Error: /' | tee -a ${errorfile}
     fi
-    
+
     exit ${exitstatus}
 fi
 set -e
--- a/remote_branch_checker/remote_branch_checker.sh
+++ b/remote_branch_checker/remote_branch_checker.sh
@@ -249,6 +249,13 @@ ${gitcmd} merge -q --no-edit FETCH_HEAD
 exitstatus=${PIPESTATUS[0]}
 if [[ ${exitstatus} -ne 0 ]]; then
     echo "Error: The ${branch} branch at ${remote} does not apply clean to ${baseref}" | tee -a ${errorfile}
+
+    mergeconflicts="$( ${gitcmd} diff --name-only --diff-filter=U )"
+    if [[ -n "${mergeconflicts}" ]]; then
+        echo "Error: Merge conflict(s) in file(s):" | tee -a ${errorfile}
+        echo "${mergeconflicts}" | sed 's/^/Error: /' | tee -a ${errorfile}
+    fi
+    
     exit ${exitstatus}
 fi
 set -e
--- a/tests/remote_branch_checker.bats
+++ b/tests/remote_branch_checker.bats
@@ -9,6 +9,11 @@ setup () {
     export phpcsstandard=$LOCAL_CI_TESTS_PHPCS_DIR
 }
 
+teardown () {
+    # Some test may end with the repo in rebase/merging conflicts status. Let's reset/abort that always.
+    cd ${gitdir} && ${gitcmd} reset --hard HEAD
+}
+
 # Assert prechecker results
 # usage: assert_prechecker branch issue commit summary
 # Performs a prechecker run and asserts the results against a collection of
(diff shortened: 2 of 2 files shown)


remote_branch_checker/remote_branch_checker.sh
```diff
@@ -249,6 +249,13 @@ ${gitcmd} merge -q --no-edit FETCH_HEAD
 exitstatus=${PIPESTATUS[0]}
 if [[ ${exitstatus} -ne 0 ]]; then
     echo "Error: The ${branch} branch at ${remote} does not apply clean to ${baseref}" | tee -a ${errorfile}
+
+    mergeconflicts="$( ${gitcmd} diff --name-only --diff-filter=U )"
+    if [[ -n "${mergeconflicts}" ]]; then
+        echo "Error: Merge conflict(s) in:" | tee -a ${errorfile}
+        echo "${mergeconflicts}" | tee -a ${errorfile}
```
stronk7, line 256: Unless I'm wrong (we don't have any test covering exactly this), I think that the value of the `${mergeconflicts}` variable must be added to the previous line (the `Error: Merge conflict(s) in:` one).

The reason is that the job that is in charge of reporting the results to the tracker only reports lines beginning by `(Error|Warn)`, so the current patch won't show the list of conflicting files. By adding the list to the previous line (that begins with `Error`) it will be reported.

Ciao :-)

stronk7 (changes requested):
Good one @ziegenberg . Only an issue that I can imagine (as commented in the patch), towards ensuring that the list of conflicting files is reported.

remote_branch_checker/remote_branch_checker.sh
```diff
@@ -249,6 +249,13 @@ ${gitcmd} merge -q --no-edit FETCH_HEAD
 exitstatus=${PIPESTATUS[0]}
 if [[ ${exitstatus} -ne 0 ]]; then
     echo "Error: The ${branch} branch at ${remote} does not apply clean to ${baseref}" | tee -a ${errorfile}
+
+    mergeconflicts="$( ${gitcmd} diff --name-only --diff-filter=U )"
```
stronk7, line 253: Tiny comment here... maybe we also need to add the `--relative` option to avoid any full path to be printed?

stronk7 (changes requested):

remote_branch_checker/remote_branch_checker.sh
ziegenberg: I don't think we need `--relative` as this is for when we run from a project subdirectory. With `--relative`, it can be said to exclude changes outside the directory and show pathnames relative to it. All other file paths are always relative to the git project. I can't recall ever seeing absolute file paths with any git command output.

remote_branch_checker/remote_branch_checker.sh
ziegenberg: If I'm not wrong, the output of `git diff --name-only --diff-filter=U` produces multi-line content if there's more than one file having merge conflicts. So, adding it to the previous line would probably only work in some cases, and we would only see the first conflicting line. Also having tabular output for a file list would be great.

How about:

```bash
echo "${mergeconflicts}" | sed 's/^/Error: /' | tee -a ${errorfile}
```

ziegenberg:
Regarding test coverage, I couldn't come up with a non-stupid test to reasonably test this scenario. Any ideas?

remote_branch_checker/remote_branch_checker.sh
stronk7: That's ok for me, I just was imagining that it would be easier to print all them in the same line and forget say with `echo "${mergeconflicts}" | xargs` or some other similar trick (`paste`...). But, as said, that's ok to me too.

stronk7:
> Regarding test coverage, I wasn't able come up with any non-stupid test to reasonably test this scenario. Any ideas?

Let me check that (tomorrow). I think we have some other test checking for merge conflicts and it's based on existing branch elsewhere. Sure we can reuse it to get this patch covered. Will share once available.

remote_branch_checker/remote_branch_checker.sh
ziegenberg: I added a fix-up commit.

stronk7:
Hi @ziegenberg ,

as promised, I've sent one extra commit to your branch that covers the patch with unit tests. It's passing here.

So, I'd suggest to just squash your 2 commits, prepare the final branch and I think this nice addition can safely land.

Ciao :-)

ziegenberg:
Hi @stronk7,

thank you for the extra unit test. I squashed and force-pushed. It's ready for merging.

ziegenberg:
Hi @stronk7, what's up with this funky gherkin unit test? I'm pretty sure I did not touch anything there...

stronk7 (approved):

PR #309, opened 2024-10-03
Ignore whitespace issues in upgrade notes yml

andrewnicols:


sarjona (approved):
Hi @andrewnicols!
Thanks a lot for working on this improvement. The patch looks good and it's great you covered with tests. Well done!

There are a couple of tests failing, but they are unrelated to the patch, so I'm going to merge this (and then, in a separate issue, we can try to fix them).

Thanks @stronk7 for your help!! :-*

PR #312, opened 2024-10-10
Post ToBiC and CiBoT results in the dedicated fields in the Tracker 

junpataleta:
* CiBoT results will be posted in the "Pre-check results" field.
* ToBiC results will be posted in the "Automated test results" field.
* The queries for bulk pre-launch jobs for awaiting IR/CLR have been updated to check for issues updated by ToBiC `$schedulemins`  (default 120 mins) since the issue was sent to CLR/IR.
* A build time has also been added to the Bulk pre-launch jobs.


junpataleta:
I updated the path so the change will initially apply to `list_of_mdls` jobs. Once tested that it works okay, we can update the other jobs to follow suit.

HuongNV13 (approved):

HuongNV13:
Thanks, Jun.
It looks good to me

PR #316, opened 2024-11-05
Acknowledge thirdpartylibs.xml in plugin names with 'work'

vmdef:
When removing files irrelevant to the tests, any files or folders containing the substring "work" remain untouched. This poses a problem if the plugin uses that substring.

See https://tracker.moodle.org/browse/MDLSITE-7849 for more details.


junpataleta:
Hi, @vmdef .

The change seems like it will ignore folders that start with `work`. Do we want to perform an exact match for the `work` folder instead? I think the pattern `^work/` suggested by @stronk7 works well for this.

Thoughts?

junpataleta:
Thanks, @vmdef! Merged!

PR #322, opened 2024-11-19
Accept any _testcase as valid

andrewnicols:
Historically we had a blur between `_test` and `_testcase`, but in recent versions we have largely solved this. We now expect:
- every test class name to be suffixed with `_test`
- every test class to be final
- every test class to extend a testcase with class name suffixed `_testcase`
- every testcase class to be abstract

Given these parameters, which have been in place for some time and are checked with phpcs, we can move away from a hardcoded list of accepted testcases, and move towards a regular expression based on the name.

### DIFF
--- a/verify_phpunit_xml/verify_phpunit_xml.sh
+++ b/verify_phpunit_xml/verify_phpunit_xml.sh
@@ -44,36 +44,18 @@ existingtests=$(cd ${gitdir} && find . -name tests | sed 's/^\.\/\(.*\)$/\1/g')
 ignoretests=""
 
 # Unit test classes to look for with each file (must be 1 and only 1). MDLSITE-2096
-# TODO: Some day replace this with the list of abstract classes, using some classmap going up to phpunit top class.
+# TODO: When we stop supporting these class names, we can remove them.
 unittestclasses="
-    advanced_testcase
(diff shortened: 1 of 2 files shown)


verify_phpunit_xml/verify_phpunit_xml.sh
```diff
@@ -44,36 +44,18 @@ existingtests=$(cd ${gitdir} && find . -name tests | sed 's/^\.\/\(.*\)$/\1/g')
...
-    core_reportbuilder_testcase
     data_loading_method_test_base
-    data_privacy_testcase
-    database_driver_testcase
-    externallib_advanced_testcase
-    googledocs_content_testcase
-    grade_base_testcase
-    lti_advantage_testcase
     manage_category_test_base
     messagelib_test
```
junpataleta, line 54: It looks like we can remove the following, and the test will still pass:
- `messagelib_test`
- `mod_quiz\\\\attempt_walkthrough_from_csv_test`

Is there a reason they were not removed here?

stronk7 (approved):
LGTM, I imagine that all remaining base classes are because we still aren’t completely free from them, specially in old branches. Cannot wait to have all them under the _testcase unique umbrella.

PR #326, opened 2025-02-07
Report list of modified upgrade.txt or UPGRADING.md files

ziegenberg:
When we experience modifications in any upgrade.txt or UPGRADING.md files it would be nice to have more information what files are affected.

We would also need a modification in file `lib/guzzlehttp/guzzle/UPGRADING.md` on the branch `local_ci_fixture_upgrade_txt_for_405` for the tests to succeed.


ziegenberg:
I don't feel particularly responsible for the ci failures of the `verify_phpuni_xml` test.

Issue #327, opened 2025-02-09
Code coverage tests are failing because package `kcov` is missing in Ubuntu 24.04


ziegenberg:
According to this [launchpad discussion](https://answers.launchpad.net/ubuntu/+source/kcov/+question/818841) the package was not and will not be packaged for Ubuntu 24.04. It's in 22.04, and it may be back in Ubuntu 25.04.

We should switch back to Ubuntu 22.04. for Code coverage tests until the package is back. Or we build it once per CI run and use it in subsequent jobs.

PR #328, opened 2025-02-09
Use Ubuntu 22.04 for code coverage tests

ziegenberg:
Ubuntu 24.04 is missing the kcov package. We can switch back to ubuntu-latest, once it points to Ubuntu 26.04

Fixes: #327


junpataleta (approved):
Thanks, Daniel. I also encountered this and raised a duplicate without realising this was already raised. Looks good to me.

PR #329, opened 2025-02-27
Handling of issues for the CLR queue during the code freeze period.

junpataleta:
* Update the `Manage queues on continuous job` so that improvements and new features in the CLR queue will not be held during the freeze period until the last week before the release (behaviour A1).

Note: Keeping the behaviour for
- A3b (last week before release, all issues unrelated to the release will be held)
- B1a (during the on-sync period, only bugs will be integrated)

* During the freeze, improvements and new features will be excluded from being triaged for the CLR queue so that they will remain in the candidate queue and be held accordingly.


HuongNV13 (approved):

HuongNV13:
Thanks, Jun.
It looks good to me.
It seems like the failing checks are not related to this patch.
I'm happy to merge this one now!

PR #331, opened 2025-02-27
Use ubuntu jammy for tests that require kcov

junpataleta:
With the `ubuntu-latest` now using `noble numbat`, the `kcov` package seems no longer available, so tests that require it are now failing. I'm not sure what the alternate would be, but pinning runs that require `kcov` to use ubuntu-jammy may be a quick fix until we find more time to look into it.


paulholden:
Is this a duplicate of https://github.com/moodlehq/moodle-local_ci/pull/328 ?

PR #332, opened 2025-03-07
Avoid holding issues just bumped from CLR to integration

mickhawkins:
This should prevent issues being integration held if they were simply bumped from CLR back to integration during the freeze. It appears the task is run half hourly, so I set a 2 hour buffer just to cover any cases where it's delayed/doesn't run for some reason.

This is my first contribution to this repo, so please let me know if anything else is required (eg specific testing requirements).


junpataleta (approved):
Thanks, @mickhawkins. The patch looks good to me. I just made some minor amendments to your commit, capitalising the first letter of the statuses.

PR #335, opened 2025-03-14
Add support for BEHAT_INIT_ARGS parameter

lameze:


sarjona (approved):
Hi @lameze!
Thanks for working on this. The patch looks good and I've been able to confirm that the 3 GH actions are consistently failing in the main branch too (without this patch), so I think it's ok to fix them in a separate issue.
I'm happy to approve this.

Should I also merge it or is that anything that you'll do? :-)

HuongNV13:
Thanks, Simey, for working on this one.
The patch looks right to me. The failures seem unrelated so I think we can address them later.

PR #337, opened 2025-04-07
Prevent incorrect holds

junpataleta:
This PR addresses two issues:
- Because security issues are not treated as "important," they are not automatically moved to the current queue. Thus, they get held, especially during the last week before the release, which is incorrect because security issues are considered related to the release. We must again treat security issues in the continuous queues manager as important.
- Must-fix and `mdlqa` issues in CLR are being incorrectly held because the query in `run_A3b` for the CLR part (`filter=23329`) does not exclude must-fix and `mdlqa` issues from being held for issues in the CLR queue.

### DIFF
--- a/tracker_automations/continuous_manage_queues/lib.sh
+++ b/tracker_automations/continuous_manage_queues/lib.sh
@@ -184,16 +184,17 @@ function run_A3a() {
 
 # A3b, add the "integration_held" + standard comment to any issue arriving to candidates (IR & CLR).
 function run_A3b() {
-    # Get the list of issues in the candidates queues (IR & CLR). All them will be held with last week comment.
+    # Get the list of issues in the candidates queues (IR & CLR) and hold with the last week comment, if necessary.
+    # All issues in CLR that don't belong to any criteria below will be held:
+    # - Must-fix issues
+    # - mdlqa issues
     ${basereq} --action getIssueList \
-               --jql "filter=14000 OR
-                  (
-                    filter=23329
-                    AND NOT (
-                      filter = 21363 OR
-                      labels IN (mdlqa)
-                    )
-                  )" \
+               --jql "filter=14000 OR (
+                      filter=23329 AND NOT (
+                        filter = 21363 OR
+                        labels IN (mdlqa)
+                      )
+                    )" \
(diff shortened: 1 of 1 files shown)


stronk7:
Hi, nice ping-pong we have here,

just for a bit of history, we used to move consider security issues as important. But it was decided that we shouldn't because the "important" classification was too much arbitrary, and manual handling was ok (sort of summary).

See https://tracker.moodle.org/browse/MDLSITE-7296 for reference.

Then, in 76ac0e78cb561b5734255fe83bd45cbe1ee1ebc1, we detected that some of the automatisms were still needed and we re-introduced them (partially). The security flagged issues was one of the conditions that was left out. Again, the same MDLSITE has it documented.

So, now, some more criteria are being re-introduced. It looks ok for me. I just would suggest to compare it with how it was working originally, mainly to see if there is any other condition to check (I remember that we were also checking privacy or testing related issues, for example).

Said that, adding back the level condition (aka security), looks ok to me.

Ciao :-)

stronk7:
And, about the 2nd part of the issue... unless I'm wrong... A2 happens before any A3, so the conditions shouldn't be needed there is the previous point already has moved the issues.

I think that, historically, that's the reason of A3b not having extra conditions, because any move should have happened already (in A2) and all the remaining issues are candidates to be held.

Ciao :-)

junpataleta:
Thanks, Eloy. Yeah, however, A2 does not handle issues in CLR already, so A3b will hold the must-fix and mdlqa CLR issues as well.

Regarding adding back the additional components (criterion `e`), the components `Privacy`, `Automated tests (Behat)`, and `PHPUnit` are now normally under CLR conditions. So, adding them back to `A2` means that they will be moved to the current queue if the CLR/IR triage job has not got to them first. Adding them to `A3b` will prevent them from being held, but they can get stuck in the IR candidates queue, especially if there are no CLRers available.

I think we should reconsider adding criterion `e` on a separate issue.

stronk7 (approved):
All ok, for me, then. Good job!

junpataleta:
Awesome. Thanks, Eloy! Self-merging this!

PR #340, opened 2025-04-11
Ignore renamed files for mustache lint

andrewnicols:
If files are only renamed, and do not contain any other changes, there is not need to mustache lint them.

This issue has come about because we're renaming all of the mustache templates in MDL-83424 and it's timing out the job.


junpataleta (approved):
Thanks, Andrew. This makes sense.

PR #341, opened 2025-04-11
Fix broken unit test

andrewnicols:
The phpunit unit test was assuming that the privacy subsystem had no tests. This was failing because there are tests.

Changing to a component which (currently) has no tests.


junpataleta (approved):
Looks good. Thanks, Andrew!

PR #342, opened 2025-04-17
Log into Docker Hub

andrewnicols:
Docker have implemented rate limiting that this repository hits.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Login to Docker Hub
         uses: docker/login-action@v3
         with:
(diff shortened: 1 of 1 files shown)


andrewnicols:
I've added an oragnisation level pair of secrets for this with read-only access to our dockerhub creds. We'll hit this in other repos.

andrewnicols:
https://www.docker.com/blog/revisiting-docker-hub-policies-prioritizing-developer-experience/

andrewnicols:
Okay, so the secrets are not available if the run is triggered from a fork. Not sure how to get around this because the new rate limit is killing us.

PR #343, opened 2025-05-15
Migrate all scripts to use standard variables for Jira

andrewnicols:


abgreeve (approved):
All the files look correctly updated.
The custom fields and filters are conveniently in one place.
Thank you.

PR #344, opened 2025-05-15
fix: mustache: allow certain non body level elements as root in linter

Menrath:
Problem was already mentioned 8 years ago. https://moodle.org/mod/forum/discuss.php?d=361491

I again stumbled into it, when I added a template for a list item because wanted to use the JavaScript to render new list items via the `Templates.renderForPromise` function.

### DIFF
--- a/mustache_lint/mustache_lint.php
+++ b/mustache_lint/mustache_lint.php
@@ -234,10 +234,38 @@ function print_message($severity, $mesage) {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


Menrath:
An suggestion: I can also offer implementing this in a optional way, so that moodle-plugin-ci may get an additional syntax in cli to activate this behaviour, so the current default behaviour stays the same.

PR #350, opened 2025-06-18
Replace tracker images with equivalent emojis.

paulholden:
There are styling problems by using static image files from the CDN as reported in https://moodle.atlassian.net/issues/ICT-5424


sarjona (approved):
Thanks a lot for working on this Paul! I love the emojis you added to replace the images :-)

PR #351, opened 2025-06-18
Separate pre-checker coloured text from link content.

paulholden:
Restores the previous lost styling to indicate pass/warn/error details as reported at https://moodle.atlassian.net/browse/ICT-5408


sarjona (approved):
Hi @paulholden!
Thanks for preparing this patch! I really miss these colors (because they help to quickly identify the errors). 
Although there is a GHA job failing, I think it can be ignored (it seems random and unrelated to this patch).

PR #352, opened 2025-06-18
Include job badge with URL in automated test results.

paulholden:
Restores previous functionality lost during migration, as reported in https://moodle.atlassian.net/issues/ICT-5409


sarjona (approved):
Hi @paulholden!
Thanks a lot forpreparing a fix to improve the test results and add the missing badge <3
The patch looks good!

PR #353, opened 2025-06-19
Re-format the job badge/URL display in automated results.

paulholden:
Follows 9f5bc544, and works around current Jira limitation in image hyperlinks and sizing, from original https://github.com/moodlehq/moodle-local_ci/pull/352


sarjona (approved):
Thanks for improving how badges are displayed! The new size looks great! <3

PR #355, opened 2025-06-24
Tag CLRs by AccountIDs

HuongNV13:
Jira Cloud does not support mentioning users directly by email address due to privacy and GDPR-related restrictions. Instead, we need to mention users using their accountId.


junpataleta (approved):
Thanks for tracking this down, Huong. The changes make sense and look good to me.

PR #359, opened 2025-07-14
Support public directory for checks (MDL-83424)

andrewnicols:
This patchset:
* makes the `list_valid_components` tooling work with the new `public` directory;
* makes the `create_phpunit_xml.php` tooling work with the public directory; and
* makes the version check tooling work too.


HuongNV13 (approved):

HuongNV13:
Thanks, Andrew

It looks good to me.

PR #368, opened 2025-12-02
remove phpcompatibility dependency

lameze:
Removing PHPCompatibility dependence from local_ci since this tool has been removed from moodle-cs.


andrewnicols:
Sorry - didn't see this when I raised https://github.com/moodlehq/moodle-local_ci/pull/369

We need to also bump the dependency on moodle-cs at the same time

PR #370, opened 2026-02-11
Only load `\phpunit_util` if not autoloadable

andrewnicols:
From Moodle 5.2 the `\phpunit_util` class has been renamed to `\core\test\phpunit\phpunit_util`, supports the autoloader, and can still be autoloaded from its old location.

Attempting to manually load the file leads to a failure.


abgreeve (approved):
Looks like we did this before with the core_renderer. Thanks for the update.

PR #372, opened 2026-03-19
Ensure CLR candidates are also held during freeze

mihailges:


lameze:
Thanks Mihail, the patch looks good to me.

From what I can see filter_candidatesForCLR is defined in moodle-local_ci/jira.sh

I really hope that we can unify the waiting for CLR/IR status to avoid having to maintain two statuses.

Happy to get this merged now.

PR #374, opened 2026-06-18
Stop applying @skip_chrome_zerosize tag to behat runs

lameze:
Stop applying @skip_chrome_zerosize tags to Behat in favor of the @skip  tag used by https://github.com/moodlehq/moodle-skip-tag-injector


lameze:
Self merging since I got no feedback on the team chat

